PR #20, opened 2018-07-05
Travis build

### LINKED ISSUE #13: set up CI
the sooner the better, especially with work being done on integration with other services.

ableuler:
Add automated testing and image building.
Integrate the gateway into renku helm deployment.
Closes #13


helm-chart/renku-gateway/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
leafty: use a symbolic link? see renku-ui

helm-chart/renku-gateway/templates/_helpers.tpl
```diff
@@ -0,0 +1,43 @@
+{{/* vim: set filetype=mustache: */}}
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "renku-gateway.name" -}}
```
leafty: Use `gateway` to stay consistent with the other sub-charts ;)

```
define "gateway.name"
...
```

helm-chart/renku-gateway/LICENSE
ableuler: good point - done!

helm-chart/renku-gateway/templates/_helpers.tpl
ableuler: Actually, I was following the ui example (see https://github.com/SwissDataScienceCenter/renku-ui/blob/master/helm-chart/renku-ui/templates/_helpers.tpl). Should I change it there too?

leafty: yes, make an issue, but it's not urgent ;)

3C111:
I ran it against the UI (gateway-integration branches) and it seemed to work fine. Did you manage to deploy it or is that not part of this PR?

PR #22, opened 2018-07-11
Fix header swapping and forwarding back

ebiiii:


app/gateway/proxy.py
```diff
@@ -80,35 +81,39 @@ def pass_through(path):
     # TODO: This switch should be resource dependent.
     if path.startswith('storage'):
         url = app.config['RENKU_ENDPOINT'] + "/api/" + path
+        auth_headers = headers
     else:
         url = app.config['GITLAB_URL'] + "/api/" + path
```
3C111: Why is this part moved in the 'else' clause? Is there no need for authentication on the storage part?

ebiiii:
I close this one and start a new branch on it to implement the generic configuration for endpoints.

PR #24, opened 2018-07-23
Generic config

ebiiii:
This makes the authentication and processing of the requests pluggable. It may however conflict with #20 , I can rebase if necessary or solve the conflicts after the other is merged.


3C111:
First: thanks so much for adding tests :] 

I tried to run against the gateway-integration-renku from the Renku UI and both the  integrate-gateway and integrate-proxy from renku, but in both cases i cannot login; I get a 404. How did you run this?

ableuler:
@3C111 I think before testing this PR, the corresponding branch has to be rebased and some conflicts have to be fixed. I think it's easiest to do this once my PRs on renku/renku-ui have been merged.

ebiiii:
I resolved the conflicts and fixed it to run the unit tests. @3C111 You can go ahead with integration tests.

3C111:
The gateway works on master with Renku (branch integrate-proxy) and Renku-UI (branch gateway-integration-renku)  When running the generic-config branch against Renku and Renku-UI I get 404 errors and it is not possible to login.

> "GET /v4//user HTTP/1.1" 404 -
>  "GET /v4//projects?starred=true HTTP/1.1" 404 -

ableuler:
I see that the project parsers that @3C111 has started to write are used here. In the master branch they are not and there need to be some changes in the UI to match the adapted project structure. I think we should takle this using different API versions. So v0 of the API gateway just forwards the GitLab entities unmodified. At the same time we can develop v1 which parses the entities given the desired Renku schema. In the UI I can then decide when I want to switch from using v0 to v1.

I suggest to deactivate the parsers for this PR and to tackle API versioning in a new issue asap.

ebiiii:
@ableuler thanks for noticing that. I was sure the master branch already used the project mapper and I was struggling to understand what was going wrong with the UI. Now it works fine (tested with minikube and telepresence).

PR #25, opened 2018-07-26
Fix urlencoded filepaths as resource identifiers in request urls

ableuler:
GitLab expects filepaths to be URL-encoded when requesting repository files. Fix the automatic unencoding done by flask.

### DIFF
--- a/app/gateway/proxy.py
+++ b/app/gateway/proxy.py
@@ -20,8 +20,9 @@
 import logging
 import json
 import requests
+import re
 from flask import request, Response
-from urllib.parse import urljoin
+from urllib.parse import urljoin, quote_plus
 
 from app.helpers.gitlab_parsers import parse_project
 from .. import app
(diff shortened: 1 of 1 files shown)


app/gateway/proxy.py
```diff
@@ -79,33 +80,39 @@ def healthcheck():
     return Response(json.dumps("Up and running"), status=200)
```
3C111: Why is it commented out? For later?

app/gateway/proxy.py
```diff
@@ -137,13 +144,41 @@ def pass_through(path, headers=None):
...
+
 @app.route(urljoin(app.config['SERVICE_PREFIX'], 'dummy'), methods=['GET'])
 def dummy():
     return 'Dummy works'
 
 
 def generate(response):
     for c in response.iter_lines():
-        logger.debug(c)
+        # logger.debug(c)
```
3C111: I think you can remove this

app/gateway/proxy.py
ableuler: It's not used at the moment, so I thought we should at least make this obvious.

PR #26, opened 2018-07-28
Match gitlab users

ableuler:
Address the problems found during deployment on testing.datascience.ch.


leafty:
Let's not forget to publish the chart

PR #28, opened 2018-08-06
Use telepresence for local development

ableuler:
Make dev setting easily accessible through `make dev`.

### DIFF
--- a/run-telepresence.sh
+++ b/run-telepresence.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+#
+# Copyright 2018 - Swiss Data Science Center (SDSC)
+# A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+# Eidgenössische Technische Hochschule Zürich (ETHZ).
+#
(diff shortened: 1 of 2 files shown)


run-telepresence.sh
```diff
@@ -0,0 +1,36 @@
...
+CURRENT_CONTEXT=`kubectl config current-context`
+
+echo "You are going to exchange k8s deployments using the following context: ${CURRENT_CONTEXT}"
+read -p "Do you want to proceed? [y/n]"
+if [[ ! $REPLY =~ ^[Yy]$ ]]
+then
+    exit 1
+fi
+FLASK_DEBUG=1 \
+    HOST_NAME=http://${MINIKUBE_IP} \
```
rokroskar: you want this indent?

ableuler: I do whatever my code editor thinks is good practice 😬 
Can remove them.

rokroskar: Oh I don't care one way or another, was just curious. I suppose the indent helps readability.

PR #43, opened 2018-08-18
Clean up Dockerfile and switch to using pipenv also during image build.

ableuler:


ableuler:
@rokroskar good point. changed!

PR #45, opened 2018-08-21
Unset specific package versions in Pipfile and recreate lockfile.

ableuler:


Dockerfile
```diff
@@ -7,7 +7,7 @@ WORKDIR /code
 RUN apt-get update && apt-get install -y gcc && \
     pip install --upgrade pip && \
     pip install pipenv && \
-    pipenv install --system
+    pipenv install --system --ignore-pipfile
```
leafty: Why not `--deploy`?

ableuler: Do we want it to fail (--dev) or just use the lockfile and ignore the pipfile (--ignore-pipfile) in case pipfile and lockfile are out of sync? The default pipenv behaviour seems to be using "*" for all the packages, so I thought ignoring the pipfile would be reasonable...

leafty: If the pipfile and lock are out of sync, it's no good. This happens only if you manually edit the pipfile.

leafty: Also, --deploy uses the lock file. That's why we had some chartpress issues only in UI and not in root

ableuler: > If the pipfile and lock are out of sync, it's no good. This happens only if you manually edit the pipfile.
Won't pipenv also try to update the lockfile whenever a new version of a "*"-dependency is released?

> Also, --deploy uses the lock file. That's why we had some chartpress issues only in UI and not in root
That's exactly what --ignore-pipfile does too, I agree we need one or the other option. Anyway, you seem to have a stronger opinion about one or the other than I have, so I'll switch to --dev ;-)

leafty: That's the point of the lock file. It avoids installing untested environments. If new packages are released, you have to recreate the lock file to use them.

PR #47, opened 2018-08-23
Manually start dev server inside telepresence

ableuler:
Killing telepresence can leave the swapped k8s deployment broken. We therefore run a shell and manually start the dev server which can be killed followed by gracefully exiting telepresence.


rokroskar:
🤣 I didn't read it but clearly I should have.

PR #50, opened 2018-08-29
Replace Flask by Quart

ebiiii:
tests are broken, it seems it is not possible to use responses and asyncio simultaneously

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,6 @@ RUN apt-get update && apt-get install -y gcc && \
 #       Currently, the poor mans approach to session handling and the
 #       signaling through blinker do only work within one process!
 
-CMD ["gunicorn", "-b 0.0.0.0:5000", "run:app.app",  "-k gevent"]
+CMD ["hypercorn", "-b", "0.0.0.0:5000", "run:app.app"]
 
 EXPOSE 5000
--- a/Pipfile
+++ b/Pipfile
@@ -4,21 +4,21 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-greenlet = "*"
+Werkzeug = "*"
+hypercorn = "*"
 oic = "*"
 requests = "*"
-responses = "*"
 blinker = "*"
-Flask = "*"
-Flask-Cors = "*"
+quart = "*"
+Quart-CORS = "*"
 PyJWT = "*"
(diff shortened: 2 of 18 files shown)


rokroskar:
https://github.com/pnuckowski/aioresponses?

ebiiii:
If we switch to this one we would also have to replace all the "requests" usages with "aiohttp". Maybe later. Here this issue is that tests are run asynchonously and the configuration changes between tests, I need to make them synchonous.

leafty:
You can't make a new event loop per test?

ebiiii:
Yes, this is the way to do it.

Dockerfile
```diff
@@ -21,6 +21,6 @@ COPY endpoints.json .
 #       Currently, the poor mans approach to session handling and the
 #       signaling through blinker do only work within one process!
 
-CMD ["gunicorn", "-b 0.0.0.0:5000", "run:app.app",  "-k gevent"]
+CMD ["hypercorn", "-b0.0.0.0:5000", "run:app.app"]
```
leafty: :unicorn: with multiple horns? Parallel thrusting ftw :rainbow:

Pipfile
```diff
@@ -4,16 +4,18 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-greenlet = "*"
+Werkzeug = "*"
+hypercorn = "*"
 oic = "*"
 requests = "*"
+pytest = "*"
```
ableuler: I think these 3 packages could go to dev-packages.

helm-chart/renku-gateway/Chart.yaml
```diff
@@ -3,4 +3,4 @@ appVersion: '1.0'
 description: A Helm chart for the Renku gateway
 name: renku-gateway
 icon: https://github.com/SwissDataScienceCenter/renku-sphinx-theme/raw/master/renku_sphinx_theme/static/favicon.png
-version: 0.1.0-a4fb826
+version: 0.0-unclean-cdf747b
```
ableuler: The chartpress relics shouldn't be committed - please revert to `version: 0.1.0` while you're at it...

ableuler:
@ebiiii I'd say we merge this to master now that the 0.2.0 release has been tagged?

PR #52, opened 2018-09-05
Refactor token handling

### LINKED ISSUE #51: Refactor token handling
Implement the server-side changes to match https://github.com/SwissDataScienceCenter/renku-ui/issues/307:

- [x] Set access tokens as secure HttpOnly cookies
- [x] Allow the access token to be sent as cookie header but enforce the presence of a custom header in this case to ensure the request was initiated by JS.
- [x] Check the Host/Referrer header and compare to the allowed origins specified in the access token
- [x] Properly set the Access-Control-Allow-Origin header

ebiiii:
~~partially~~ fixes #51  ~~(the last point still need to be done)~~

--
Do not merge before https://github.com/SwissDataScienceCenter/renku-ui/issues/307 is resolved too!

### DIFF
--- a/app/auth/keycloak_auth.py
+++ b/app/auth/keycloak_auth.py
@@ -16,7 +16,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
-from .web import get_refreshed_tokens
 
 logger = logging.getLogger(__name__)
 
@@ -24,9 +23,5 @@ logger = logging.getLogger(__name__)
 class KeycloakAccessToken():
 
     def process(self, request, headers):
-
-        new_tokens = get_refreshed_tokens(headers)
-        if new_tokens:
-            headers['Authorization'] = "Bearer {}".format(new_tokens.get('access_token'))
 
         return headers
--- a/app/auth/web.py
+++ b/app/auth/web.py
@@ -21,6 +21,8 @@ import json
 import time
 import logging
 import re
+import os
+
 from oic.oauth2.grant import Token
 from quart import request, redirect, url_for, current_app, Response
 from urllib.parse import urljoin
@@ -55,6 +57,11 @@ if 'gateway.renku.build' in app.config['HOST_NAME']:
 else:
     COOKIE_DOMAIN = None
 
+if app.config['HOST_NAME'].startswith("https"):
+    COOKIE_SECURE = {'secure': True, 'httponly': True}
+else:
+    COOKIE_SECURE = {'httponly': True}
+
 # We use a short-lived dictionary to store ongoing login sessions.
 # This should not grow in size and can easily be trashed when the service needs
 # to be restarted.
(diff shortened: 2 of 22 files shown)


app/auth/keycloak_auth.py
```diff
@@ -25,8 +24,4 @@ class KeycloakAccessToken():
 
     def process(self, request, headers):
```
3C111: So now this, does nothing?

ebiiii: As keycloak tokens are already present in the headers (used by the UI too), this one has nothing to do, yes.

3C111: But you think we can find a use for this function? Else we might just refactor it out.

app/gateway/proxy.py
```diff
@@ -16,43 +16,45 @@
...
 import json
 import jwt
-from flask import request, Response
+import re
+from quart import request, Response
 from urllib.parse import urljoin
 
 from .. import app
-from flask_cors import CORS
+from app.auth.web import COOKIE_DOMAIN, COOKIE_SECURE, get_refreshed_tokens
```
3C111: CORS is no longer needed?

3C111:
Tests run! Did you try it against the platform? I am getting some login errors but it might be me...

app/auth/web.py
```diff
@@ -56,6 +57,11 @@
 else:
     COOKIE_DOMAIN = None
 
+if os.environ.get('QUART_DEBUG', "0") != "1":
+    COOKIE_SECURE = {'secure': True, 'httponly': True}  # keep httponly when UI is also adapted
```
ableuler: I think this will currently not work for a minikube-deployment in which the gateway is NOT swapped by telepresence. In this case `QUART_DEBUG` will not be set, therefore `secure` will be `True`, however the UI is still not served through https.

I suggest to keep `httponly` always set (less risk of introducing bugs in a dev setting where js reads from the cookie) and the `COOKIE_SECURE` env variable should be set in `deployment.yaml` inside the chart templates (there is already a `gateway.protocol`) defined that you can use: https://github.com/SwissDataScienceCenter/renku-gateway/blob/master/helm-chart/renku-gateway/templates/_helpers.tpl

ebiiii: I set the httponly for both cases and changed the condition to look at the HOST_NAME, which is injected with the gateway.protocol and then should reflect the fact that we are using https or not.

But it seems the UI is not currently sending the X-Requested-With header

app/gateway/proxy.py
ebiiii: no, it is now at the app root level

app/auth/web.py
ableuler: Yes, the UI needs to be updated too to match the changes in the gateway.

app/gateway/proxy.py
ableuler: Actually it could be removed there too. The more restrictive approach to token handling between UI and gateway discussed [here](https://github.com/SwissDataScienceCenter/renku-ui/issues/307) relies on UI and gateway to be running under the same host name. We therefore do not do any cross origin resource sharing (CORS) anymore, and we therefore don't have to support it on the server side anymore.

app/auth/keycloak_auth.py
ebiiii: There is no superclass implementing the defaults, so the method has to exist here. We will see if it makes sense to have a base_auth class in the future.

PR #54, opened 2018-09-18
Forward the pagination headers in the GitLab response

ableuler:
Forward pagination headers in the response, modify URLs in the `Link` header such that they point to a gateway URL.

### DIFF
--- a/app/processors/gitlab_processor.py
+++ b/app/processors/gitlab_processor.py
@@ -9,6 +9,7 @@ import requests
 import re
 
 from flask import Response
+from werkzeug.datastructures import Headers
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


app/processors/gitlab_processor.py
```diff
@@ -45,8 +56,30 @@ def urlencode_paths(path):
...
+            app.config['GITLAB_URL'], app.config['HOST_NAME']
+        )
+
+    return headers
+
+
 class GitlabGeneric(BaseProcessor):
 
+    def __init__(self, path, endpoint):
+        self.path = path
```
ebiiii: why not `super().__init__(path, endpoint)`

app/processors/gitlab_processor.py
```diff
@@ -56,6 +89,17 @@ def process(self, request, headers):
         self.endpoint = urljoin(self.endpoint.format(**app.config), self.path)
         return super().process(request, headers)
 
+    def create_response_headers(self, response):
+
+        headers = Headers()
+        for header in response.headers:
+            if header in self.forwarded_headers:
+                headers.add(header, response.headers[header])
```
ebiiii: here too

app/processors/gitlab_processor.py
ableuler: yep, you're right!

app/processors/gitlab_processor.py
ableuler: here too ;)

PR #58, opened 2018-10-08
Moving to a stateful gateway

ebiiii:
- Adds redis as a session backend for storing the users' identity as KC refresh_token (making the server side session expires with the refresh token). 
- Store the various user tokens also in redis.
- It also removes the need for blinker and the signals as well as the global variables, making it possible to have several instance of the gateway (sharing the redis instance/cluster).
- TODOS:
  - [x]  merge https://github.com/SwissDataScienceCenter/renku/pull/428
  - [x]  add "{GATEWAY_URL}/api/auth/gitlab/token" and "read_repository read_registry openid" to the scopes of the renku-ui client in gitlab. -> (https://github.com/SwissDataScienceCenter/renku/pull/440)
  - [x] mock redis for the tests.
  - [ ] merge with: https://github.com/SwissDataScienceCenter/renku-ui/pull/331

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -18,6 +18,12 @@
 """Quart initialization."""
 
 import logging
+import redis
+import sys
+import quart.flask_patch
+from flask_kvsession import KVSessionExtension
+from simplekv.decorator import PrefixDecorator
+from simplekv.memory.redisstore import RedisStore
 from quart import Quart
 from quart_cors import cors
 
@@ -26,12 +32,23 @@ from .config import config, load_config
 logging.basicConfig(level=logging.DEBUG)
 
 app = Quart(__name__)
-app = cors(app)
 
 for key in config.keys():
     app.config[key] = config[key]
 
+app = cors(app, allow_headers=['X-Requested-With'], allow_origin="*")
+
 load_config()
+
+if "pytest" in sys.modules:
+    from mockredis import mock_strict_redis_client
+    store = RedisStore(mock_strict_redis_client())
+else:
+    store = RedisStore(redis.StrictRedis(host=app.config['REDIS_HOST']))
+
+prefixed_store = PrefixDecorator('sessions_', store)
+KVSessionExtension(prefixed_store, app)
 
 from .gateway import proxy
 from .auth import web
+from .auth.gitlab_auth import gitlab_login, gitlab_get_tokens, gitlab_logout
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -36,7 +36,7 @@ app = Quart(__name__)
 for key in config.keys():
     app.config[key] = config[key]
 
-app = cors(app, allow_headers=['X-Requested-With'], allow_origin="*")
+app = cors(app, allow_headers=['X-Requested-With'], allow_origin=app.config['ALLOW_ORIGIN'])
 
 load_config()
 
--- a/app/auth/web.py
+++ b/app/auth/web.py
@@ -21,8 +21,9 @@ import json
 import time
 import logging
 import re
+import urllib.parse
 from oic.oauth2.grant import Token
-from quart import request, redirect, url_for, current_app, Response
+from quart import request, redirect, url_for, current_app, Response, session
 from urllib.parse import urljoin
 
 from oic.oic import Client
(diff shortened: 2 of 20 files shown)


rokroskar:
what is the problem with merging SwissDataScienceCenter/renku#428?

leafty:
The point to solve is to give the GitLab chart the good version number and to publish it

app/__init__.py
```diff
@@ -26,12 +32,23 @@
 logging.basicConfig(level=logging.DEBUG)
 
 app = Quart(__name__)
-app = cors(app)
 
 for key in config.keys():
     app.config[key] = config[key]
 
+app = cors(app, allow_headers=['X-Requested-With'], allow_origin="*")
```
ebiiii: Allow_origin will be the placeholder for when we will need to limit the gateway access to specific domains. But I was wondering if this would prevent from having custom UI or services using also the APIs? (so I left it open for the moment with "*")

helm-chart/renku-gateway/Chart.yaml
```diff
@@ -3,4 +3,4 @@ appVersion: '1.0'
 description: A Helm chart for the Renku gateway
 name: renku-gateway
 icon: https://github.com/SwissDataScienceCenter/renku-sphinx-theme/raw/master/renku_sphinx_theme/static/favicon.png
-version: 0.2.0
+version: 0.0-unclean-fd237f2
```
ableuler: I think this should not be committed.

app/auth/web.py
```diff
@@ -49,18 +42,6 @@
 JWT_ALGORITHM = 'HS256'
```
rokroskar: unused

app/auth/web.py
```diff
@@ -80,31 +61,101 @@
...
+        if jwt.decode(m.group('token'), verify=False).get('typ') in ['Offline', 'Refresh']:
+            logger.debug("Swapping the token")
+            to = Token(resp={'refresh_token': m.group('token')})
+            token_response = client.do_access_token_refresh(token=to)
+
+            if 'access_token' in token_response:
+                try:
+                    a = jwt.decode(
+                        token_response['access_token'], app.config['OIDC_PUBLIC_KEY'],
+                        algorithms='RS256',
```
rokroskar: use `JWT_ALGORITHM`?

app/config.py
```diff
@@ -30,10 +30,29 @@
...
+
+config['ALLOW_ORIGIN'] = os.environ.get('GATEWAY_ALLOW_ORIGIN', "").split(',')
+
+config['REDIS_HOST'] = os.environ.get('GATEWAY_REDIS_HOST', 'renku-gw-redis')
+
 config['RENKU_ENDPOINT'] = os.environ.get('RENKU_ENDPOINT', 'http://renku.build')
 config['GITLAB_URL'] = os.environ.get('GITLAB_URL', 'http://gitlab.renku.build')
 config['GITLAB_PASS'] = os.environ.get('GITLAB_PASS', 'dummy-secret')
+config['GITLAB_CLIENT_ID'] = os.environ.get('GITLAB_CLIENT_ID', 'renku-ui')
+config['GITLAB_CLIENT_SECRET'] = os.environ.get('GITLAB_CLIENT_SECRET', 'no-secret-needed')
```
rokroskar: should be required

app/config.py
```diff
@@ -30,10 +30,29 @@
 
 
 config = dict()
+
 config['HOST_NAME'] = os.environ.get('HOST_NAME', 'http://gateway.renku.build')
+
+config['SECRET_KEY'] = os.environ.get('GATEWAY_SECRET_KEY', 'dummy-secret')
```
rokroskar: should be required

app/config.py
ebiiii: you mean I should `exit(1)` or raise an exception or just don't set a default value (in which case it would be set to None and fail somewhere later if Gitlab expects a client secret)

app/config.py
ebiiii: Here I have to fail immediately

app/config.py
rokroskar: Maybe it's OK to have a default so it's not annoying for development, but at least raising a warning would be good

ebiiii:
I'm closing this one in favor of #63 (which also contains the changes from here)

PR #63, opened 2018-10-22
Stateful gateway with Gitlab and JupyterHub user tokens

### LINKED ISSUE #30: Logout from all backend services
A logout should also trigger logout from all "3rd party" backend services (like Jupyterhub and GitLab).

### LINKED ISSUE #31: Store user-specific access tokens
In the longer run the gateway has to handle state, i.e. store user-specific access tokens for all backend services which do not accept a keycloak token for their API. This will allow us to integrate backend services for which we do not have admin rights (for example not sudo-token for GitLab). The tokens could for example be stored in a redis store.

### LINKED ISSUE #51: Refactor token handling

### LINKED ISSUE #59: Collect jupyterhub token in the gateway too
Once it has the redis backend.

ebiiii:
~~Follows #58 (to be merged before)~~

Must be merged with https://github.com/SwissDataScienceCenter/renku-ui/pull/331

Closes #51 
Closes #59
Closes #30 
Closes #31


rokroskar:
I'm confused, does 5cfcbad address the comments from https://github.com/SwissDataScienceCenter/renku-gateway/pull/58?

rokroskar:
I've lost track of the dependencies between the different PRs -- could you put the final combination that needs to be merged together up on staging?

README.rst
```diff
@@ -67,3 +67,19 @@ You can run tests with
 
     $ pipenv run pytest
 
+
+Login workflow
+--------------
+
+To be able to collect the user's token from the various backend services, the gateway uses the OAuth2/OIDC protocol and redirect the users to each of them.
```
rokroskar:
```suggestion
To be able to collect the user's token from the various backend services, the gateway uses the OAuth2/OIDC protocol and redirects the users to each of them.
```

app/auth/web.py
```diff
@@ -80,140 +59,243 @@
...
+    logout_url = '{}/protocol/openid-connect/logout?{}'.format(
+        app.config['OIDC_ISSUER'],
+        urllib.parse.urlencode({'redirect_uri': request.args.get('redirect_url')}),
+    )
     response = await app.make_response(redirect(logout_url))
-    response.delete_cookie('access_token', domain=COOKIE_DOMAIN)
-    response.delete_cookie('refresh_token', domain=COOKIE_DOMAIN)
-    response.delete_cookie('id_token', domain=COOKIE_DOMAIN)
+
+    a = jwt.decode(session['token'], verify=False)
```
ableuler: For me the logout is still failing here. The session does not contain a 'token', in fact it's entirely empty. However, if I check the content of the Redis store directly, the token is there. However, if I add
```
 import asyncio
 await asyncio.sleep(1)
```
at the beginning of the method, things work fine. So I assume that the Redis session lookup should be treated ("awaited") as an asynchronous call. Maybe we have to drop the convenience of the standard session implementation here and connect to Redis explicitly using something like `aioredis` (asynchronous Redis client library)?

app/__init__.py
```diff
@@ -26,12 +32,24 @@
...
 
+app = cors(app, allow_headers=['X-Requested-With'], allow_origin=app.config['ALLOW_ORIGIN'])
+
 load_config()
 
+if "pytest" in sys.modules:
+    from mockredis import mock_strict_redis_client
+    store = RedisStore(mock_strict_redis_client())
+else:
+    store = RedisStore(redis.StrictRedis(host=app.config['REDIS_HOST']))
```
ableuler: The redis store must be password protected. Currently it's very easy for a user to hack redis (in fact I tried...) 
image

ableuler: Update: This is actually not as pressing as I thought because the network policy will prevent this kind of access for non-minikube deployments (as pointed out by @leafty).

app/auth/web.py
ebiiii: Can you also try the /api/auth/user or /api/auth/info (after a successful login)? Those two endpoints do also need to read the session to get the token. (if there was a general issue about accessing the session data, the login would have failed retrieving the state, I think it is just a matter of refactoring this endpoint to do the session access in the synchronous part of the method, i.e. before the first yield/await).

rokroskar:
would be great if you could also update the API Gateway page in the main renku docs, especially the [current state section](https://renku.readthedocs.io/en/latest/developer/services/api_gateway.html#current-state-of-implementation)

rokroskar:
@ebiiii is this PR finished? @ableuler is on vacation for 10 days so if this can be tested and approved asap so we can move forward with 0.3.0 that would be great.

app/auth/web.py
ableuler: Thanks @ebiiii for calling it a "trick" while it is more of a bug in the UI ;)

ebiiii:
Yes, the gateway doesn't log out from gitlab at the moment. I don't see a simple solution right now as Keycloak doesn't agree to redirect to gitlab's logout if it is not part of the "redirect_uri" list (but we might use some other tricks). As this is not preventing Renku to work, I would suggest to open an issue and solve it after this one is merged.

ableuler:
@ebiiii But this will lead to really weird errors when you're logged in as user A, then you log out and you log in as user B and connect to gitlab, no? If it's only the logout itself which is affected, then I'm fine delaying this to later. 
>Keycloak doesn't agree to redirect to gitlab's logout if it is not part of the "redirect_uri" list 
Is there a problem with adding all gitlab routes as allowed redirect uri's to Keycloak?

helm-chart/renku-gateway/templates/secret.yaml
```diff
@@ -11,3 +11,6 @@ type: Opaque
 data:
   gitlabSudoToken: {{ .Values.gitlabSudoToken | default .Values.global.gitlab.sudoToken | b64enc | quote }}
   oidcClientSecret: {{ .Values.oidcClientSecret | default .Values.global.gateway.clientSecret | b64enc | quote }}
+  gitlabClientSecret: {{ .Values.gitlabClientSecret | default .Values.global.gateway.gitlabClientSecret | b64enc | quote }}
+  jupyterhubClientSecret: {{ required "Fill in .Values.jupyterhub.clientSecret with `openssl rand -hex 32`"  .Values.jupyterhub.clientSecret | b64enc | quote }}
+  gatewaySecret: {{ .Values.secretKey | b64enc | quote }}
```
leafty: add same guard as above
```suggestion
  gatewaySecret: {{ required "Fill in .Values.secretKey with `openssl rand -hex 32`"  .Values.secretKey | b64enc | quote }}
```

helm-chart/renku-gateway/values.yaml
```diff
@@ -35,15 +35,35 @@ replicaCount: 1
...
+## The client secret for authentication against gitlab
+# gitlabClientSecret: no-secret
+
+## You can restrict here the allowed origins for CORS.
+## This can be a single value, a comma-separeted list or *.
+# allowOrigin: "*"
+
+## For production deployment, you will need to define the secret key
+## This is a random string, used for cryptographic operations on cookies
+secretKey: dummy-secret
```
leafty: Do not provide default

helm-chart/renku-gateway/values.yaml
```diff
@@ -35,15 +35,35 @@ replicaCount: 1
...
+
+jupyterhub:
+  ## the url of the jupyterhub deployment
+  # url:
+  ## Client ID, this should match the jupyter hub config of services: oauth_client_id
+  clientId: gateway
+  ## Client secret, this should match the api_token
+  clientSecret: '' # use `openssl rand -hex 32`
```
leafty: Do not provide default

ebiiii:
@ableuler I fixed the logout (It was a mistake of mine in the URL formatting). See https://github.com/SwissDataScienceCenter/renku/pull/469 for the next issue with logout ;-)

app/auth/jupyterhub_auth.py
```diff
@@ -0,0 +1,115 @@
...
+
+    state = rndstr()
+
+    session['login_seq'] += 1
+    session['jupyterhub_state'] = state
+
+    args = {
+        'client_id': app.config['JUPYTERHUB_CLIENT_ID'],
+        'response_type': 'code',
+        'redirect_uri': app.config['HOST_NAME'] + url_for('jupyterhub_get_tokens'),
```
leafty: This needs to be under `$JUPYTERHUB_URL/$SERVICE_PREFIX`

ebiiii: this is fixed by setting ```notebooks.jupyterhub.hub.services.gateway.oauth_redirect_uri=https://<HOST_NAME>/api/auth/jupyterhub/token``` in the values when deploying renku. It is already injected with the minikube-deploy script, but must be added manually for other deployments

helm-chart/renku-gateway/Chart.yaml
```diff
@@ -3,4 +3,4 @@ appVersion: '1.0'
 description: A Helm chart for the Renku gateway
 name: renku-gateway
 icon: https://github.com/SwissDataScienceCenter/renku-sphinx-theme/raw/master/renku_sphinx_theme/static/favicon.png
-version: 0.2.0
+version: 0.2.1
```
ableuler: 0.3.0 or 0.2.1 ?

PR #70, opened 2018-11-27
Don't try to validate refresh token and use access tokens instead

ebiiii:
Keycloak 4.5.0 changed the signature of refresh tokens which are now considered as INTERNAL and this is hardcoded. ( https://github.com/keycloak/keycloak/pull/5533#discussion_r215880421). Validation is only possible through a rest endpoint, so we decided to only validate access tokens and use them for the sessions.


leafty:
Ihmo you should also store the refresh token to renew the access tokens

ableuler:
@leafty The refresh token is in fact stored and used to retrieve a new access token once the old one has expired.

rokroskar:
If this is to go in for the 0.3.0 release, please add it to the milestone (@ebiiii @ableuler @leafty )

PR #73, opened 2018-12-13
Fix request headers before forwarding

ableuler:
Remove a header from the request before forwarding which can cause problems related to ssl.

### DIFF
--- a/app/processors/base_processor.py
+++ b/app/processors/base_processor.py
@@ -1,3 +1,4 @@
+import os
 import json
 import logging
 import requests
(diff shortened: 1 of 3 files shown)


test.json
```diff
@@ -0,0 +1 @@
+{"obj(ect)?/[\\d-]+/": {"endpoint": "http://localhost/api", "processor": "app.tests.test_config.DummyProcessor"}}
```
ableuler: Hmm, why is this necessary? Did the tests not work before?

ciyer: I thought that file came from you, but I now see that it came from my commit. I've removed it.

PR #74, opened 2018-12-13
global: split to blueprints

jirikuncar:


rokroskar:
@jirikuncar can you clarify in the description what this is doing? I don't understand what blueprints have to do with SPARQL queries...

jirikuncar:
This was a quick TODO note before leaving on holidays ;)

PR #77, opened 2019-01-14
Shorten request timeouts

ableuler:
Set all request timeouts to 30 instead of 300 seconds.
Note that for streaming requests, the timeout does NOT refer to the duration of the entire response streaming but only the first byte being transferred, see: http://docs.python-requests.org/en/master/user/advanced/#timeouts


rokroskar:
should this be a value configurable via an environment variable?

ableuler:
> should this be a value configurable via an environment variable?

I thought it was better to leave it hardcoded as we might want use different timeouts for different requests.

ableuler:
More recent changes made this PR obsolete.

PR #78, opened 2019-01-20
Fix synchronous requests

ableuler:
Use asynchronous (aiohttp) requests. Note that some of the requests made during login are still synchronous. However, for communication with the gitlab or jupyterhub API, this greatly improves the performance of the gateway.

### DIFF
--- a/app/processors/base_processor.py
+++ b/app/processors/base_processor.py
@@ -1,7 +1,7 @@
 import os
 import json
 import logging
-import requests
+import aiohttp
 
 from quart import Response
 from werkzeug.datastructures import Headers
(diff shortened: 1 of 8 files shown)


app/processors/base_processor.py
```diff
@@ -36,31 +36,35 @@ def __init__(self, path, endpoint):
...
+
+                return Response(
+                    # response=response_generator(response),
+                    response=response_data,
+                    headers=self.create_response_headers(response),
+                    status=response.status
+                )
+
+    # TODO: This does not work yet.
+    # TODO: Enable streaming responses at a later point.
```
jirikuncar: maybe you want to use https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.StreamResponse.write

PR #96, opened 2019-02-21
feat(graph): return json-ld lineage data

jirikuncar:
Please check the output format.


ableuler:
Works as advertised. After discussion I think it's best to keep this PR open until we have a consumer for json-ld and then make it available instead of the plain json if the client asks for it through the header (`Accept: application/ld+json`).

jirikuncar:
Please re-open when needed.

The query needs to be updated to follow the current logic!

PR #99, opened 2019-02-28
Fix and update the `run-telpresence.sh` script

ableuler:
- Fix path to quart app
- Adapt to our dev cluster setup
- Use `inject-tcp` method per default only on non-linux machines


lorenzo-cavazzi:
After creating a virtualenv the script runs well

PR #109, opened 2019-03-21
Allow running gateway in a debugger

ciyer:


app/__init__.py
```diff
@@ -33,6 +34,18 @@
 from .blueprints import gitlab, graph, jupyterhub, notebooks, webhooks
 from .config import config
 
+# Wait for the VS Code debugger to attach if requested
+VSCODE_DEBUG = os.environ.get('VSCODE_DEBUG') == "1"
+print("{}:{}".format(os.environ.get('VSCODE_DEBUG'), VSCODE_DEBUG))
```
ableuler: This will evaluate to `None:False` if the env variable is not set.

run-telepresence.sh
```diff
@@ -53,6 +53,14 @@ QUART_DEBUG=1 \
 QUART_APP=app:app \
 PYTHONASYNCIODEBUG=1 \
 ${QUART_EXECUTABLE} run"
+echo ""
```
ableuler: I think just adding a hint that one can activate virtual studio debugging by adding `VSCODE_DEBUG=1` to the command is enough, no need to give the full command yet again. It's a quite specific tool for a quite specific code editor.

app/__init__.py
ciyer: Sorry, that was some left-over logging. Thanks for fixing the problem with the tests.

run-telepresence.sh
ciyer: It seems like more and more of the team is migrating to VS Code, and having access to a debugger makes it easier for someone who is not familiar with the codebase to quickly be productive. Being able to copy/paste the command without having to put it together from multiple fragments reduces one more barrier to people working on the gateway. 

I think making this as easy as possible will encourage others to fix problems in the gateway themselves, but if you find it annoying, we can obviously modify the message. 

@lorenzo-cavazzi Any thoughts?

ableuler: Ok, fine for me. As long as it's just two possible commands I don't really mind it too much. And If you thinks it will help to get other people into fixing problems on the gateway, then I'm all for it :)

PR #110, opened 2019-03-22
Authorize webhooks through bearer token

### LINKED ISSUE #107: Call to create webhook sends no or invalid oauth token
It looks at in some circumstances when Gatway does the POST to webhook-services' `/projects/:id/webhooks` it does send `OAUTH-TOKEN` header with some invalid token. We tried a few tests on the UI-team environment and when we did curl webhook-services' endpoint from its container console it works properly. So effectively these two calls where succeeding while failing with 401 from gitlab, when the same endpoint was hit by the Gateway.

`curl -XPOST --header "OAUTH-TOKEN: 07fbaa9b39f0e208ab1cf48a819888dfe8740613cbd25bc0d2cb58700872e874" http://localhost:9001/projects/72/webhooks/validation -v`

`curl -XPOST --header "OAUTH-TOKEN: 07fbaa9b39f0e208ab1cf48a819888dfe8740613cbd25bc0d2cb58700872e874" http://localhost:9001/projects/72/webhooks -v`

lorenzo-cavazzi:
Use `Authorization: Bearer xxx` Instead of `OAUTH-TOKEN: xxx` for the authentication.

Fixes #107

### DIFF
--- a/app/blueprints/webhooks.py
+++ b/app/blueprints/webhooks.py
@@ -41,5 +41,5 @@ async def forward_to_webooks(project_id, path=''):
     return await pass_through(
         request,
         processor,
-        GitlabUserToken(header_field='OAUTH-TOKEN', header_prefix='')
+        GitlabUserToken(header_field='Authorization', header_prefix='Bearer ')
     )


lorenzo-cavazzi:
I can't merge into master, waiting for someone with the proper authorization to do it.

@jachro I also wasn't, but making this very small change helped me to understand it a bit more. Btw thanks for your help in finding where the problem was!
Once this PR is merged the support for `OAUTH-TOKEN` on renku-graph for the webhook part shouldn't be necessary anymore

jachro:
Looking forward to that, then.

PR #112, opened 2019-03-27
BREAKING CHANGE: use Traefik as a proxy

jirikuncar:
Use Trafiek with a static configuration for proxing UI requests.


helm-chart/renku-gateway/templates/configmap.yaml
```diff
@@ -0,0 +1,67 @@
...
+      dashboard = true
+    [providers]
+      [providers.file]
+        watch = true
+        directory = "/config"
+    [entrypoints]
+      [entrypoints.http]
+        address = ":{{ .Values.service.port }}"
+    [http.routers]
+      [http.routers.whoami]
```
jirikuncar: this will be removed ...

helm-chart/renku-gateway/templates/configmap.yaml
```diff
@@ -0,0 +1,67 @@
...
+        method = "drr"
+        [[http.services.gitlab.LoadBalancer.servers]]
+          url = {{ .Values.gitlabUrl | default (printf "%s://%s/gitlab" (include "gateway.protocol" .) .Values.global.renku.domain) | quote }}
+          weight = 1
+      [http.services.jupyterhub.LoadBalancer]
+        method = "drr"
+        [[http.services.jupyterhub.LoadBalancer.servers]]
+          url = {{ .Values.jupyterhub.url | default (printf "%s://%s/jupyterhub" (include "gateway.protocol" .) .Values.global.renku.domain) | quote }}
+          weight = 1
+      [http.services.whoami.LoadBalancer]
```
jirikuncar: this will be removed ...

helm-chart/renku-gateway/templates/deployment.yaml
```diff
@@ -28,6 +32,75 @@ spec:
...
+  selector:
+    matchLabels:
+      app: {{ template "gateway.name" . }}-auth
+      release: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        app: {{ template "gateway.name" . }}-auth
+        release: {{ .Release.Name }}
+        {{ include "call-nested" (list . "redis" "redis.fullname") }}-client: "true"
```
jirikuncar: @ableuler I forgot to move this label. Does it work for you now?

rokroskar: did you turn off the redis network policy also?

jirikuncar: where?

rokroskar: The reason that line is there is because the redis network policy only allows connections from pods with that label afaik

rokroskar: See https://github.com/SwissDataScienceCenter/renku-gateway/blob/master/helm-chart/renku-gateway/values.yaml

jirikuncar: It was my mistake as I moved it to metadata label not pod label.

rokroskar: Aha! Sorry, I missed that part :)

ableuler: @jirikuncar No, still doesn't work. The gw still tries to connect to its own redis, but no such service is being created anymore. Just compare the services under the namespaces `andreas` and `renku` currently deployed on the dev cluster. Did you rip out the second redis on purpose?

ableuler:
@jirikuncar Just a quick info for you on the traefik 2.0 config that you put together (so I thought you might be interested still ;) :

```
      [http.middlewares.gitlab.AddPrefix]
        prefix = "/gitlab/api/v4"
```

```
      [http.services.gitlab.LoadBalancer]
        method = "drr"
        [[http.services.gitlab.LoadBalancer.servers]]
          url = "https://dev.renku.ch/gitlab/"
          weight = 1
```
If you look at the above blocks of config, you realise that this should result in  `/gitlab/gitlab`. Now traefik seems to completely ignore the entire path given in the url of a server. That’s why the above config even works. Do you think that’s desired behaviour of traefik or a bug?

jirikuncar:
@ableuler I am not sure if it is desired -- it is definitely confusing. The source of problem is a combination of service with URL prefix and custom prefix. Just look at the middleware as a piece that modifies the path which should start with the same prefix as the service.

Issue #113, opened 2019-03-29
Improve logging strategy


ableuler:
Currently, there is no coherent strategy about what should be logged in development and production mode respectively.

jirikuncar:
Most of the important logs are done by Traefik now.

rokroskar:
> Most of the important logs are done by Traefik now.
This is the hope -- however, the traefik logs seem to contain a lot of 

```
time="2019-04-16T21:38:42Z" level=error msg="Error while Peeking first byte: EOF"
```

and nothing else. Hmm...

ableuler:
@rokroskar see #125.

rokroskar:
looks better now, but still getting this polluting the traefik logs: 

```
time="2019-04-17T16:26:31Z" level=error msg="Error while Peeking first byte: EOF"
```

ableuler:
I know, feel free to create a separate issue for it.

rokroskar:
Do we need to keep this open?

ableuler:
I would like to keep it open. There are still commented out `logger.debug` statements in the auth part of the gateway code. This doesn't make sense. If those are helpful log messages for debugging, those should be activated and deactivated through the log level and not through commenting out stuff.

rokroskar:
I see! Yes that should be fixed, agreed

Issue #98, opened 2019-02-27
Solve import issues


jirikuncar:
- [ ] `import app` should work without side effects
- [ ] `import app.config` should never `sys.exit` but rather raise an exception if necessary, better to the checks in the `create_app`

PR #115, opened 2019-04-02
Fix logging

### LINKED ISSUE #106: Fix forwarding of logger.info() logging in k8s
All of the `logger.info()` logging is missing in the k8s logs while `logger.debug()` logs are present. This can lead to situations where the gateway pod is dying without leaving any logs at all.

----
Original description by @rokroskar: 
At the moment, there are no logs provided when things go wrong during bootstrapping of the gateway. For example, if keycloak isn't reachable, there is no message just an unresponsive service. This makes debugging of new deployments difficult.

ableuler:
- Remove side effects during imports
- Check for a keycloak public key before_request() if it is not part of the app context already
- Use standard quart app.logger everywhere except during module import
- Use warnings.warn() for warnings during import

This fixes the logging problems that have partially prevented logs from showing up in production.
Closes #106 and #98. A better logging strategy should be introduced later (see #113).

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -24,15 +24,16 @@ import os
 
 import quart.flask_patch
 import redis
+import requests
 from flask_kvsession import KVSessionExtension
-from quart import Quart, Response
+from quart import Quart, Response, current_app
 from quart_cors import cors
 from simplekv.decorator import PrefixDecorator
 from simplekv.memory.redisstore import RedisStore
 
 from .auth import gitlab_auth, jupyterhub_auth, web
 from .blueprints import gitlab, graph, jupyterhub, notebooks, webhooks
-from .config import config
+from app import config
 
 # Wait for the VS Code debugger to attach if requested
 VSCODE_DEBUG = os.environ.get('VSCODE_DEBUG') == "1"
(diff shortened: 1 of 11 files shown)


app/auth/gitlab_auth.py
```diff
@@ -54,7 +51,7 @@ def process(self, request, headers):
             re.IGNORECASE
         )
         if m:
-            # logger.debug('outgoing headers: {}'.format(json.dumps(headers))
+            # app.logger.debug('outgoing headers: {}'.format(json.dumps(headers))
```
jirikuncar: better `current_app`

app/auth/web.py
```diff
@@ -80,7 +80,7 @@ def get_valid_token(headers):
         if jwt.decode(
             m.group('token'), verify=False
         ).get('typ') in ['Offline', 'Refresh']:
-            logger.debug("Swapping the token")
+            app.logger.debug("Swapping the token")
```
jirikuncar: `current_app`

app/__init__.py
```diff
@@ -66,33 +68,40 @@
...
 async def healthcheck():
     return Response(json.dumps("Up and running"), status=200)
 
 
-def _join_url_prefix(*parts):
-    """Join prefixes."""
-    parts = [part.strip('/') for part in parts if part and part.strip('/')]
-    if parts:
-        return '/' + '/'.join(parts)
+def add_blueprints(app):
```
jirikuncar: if you use `current_app` you won't have a problem with circular dependencies and you can keep it on global level.

app/config.py
```diff
@@ -19,125 +19,73 @@
...
-logger = getLogger(__name__)
-
-config = dict()
-
-config['HOST_NAME'] = os.environ.get('HOST_NAME', 'http://gateway.renku.build')
+HOST_NAME = os.environ.get('HOST_NAME', 'http://gateway.renku.build')
 
 if 'GATEWAY_SECRET_KEY' not in os.environ and "pytest" not in sys.modules:
-    logger.critical(
+    warnings.warning(
```
jirikuncar:
```suggestion
    warnings.warn(
```

PR #119, opened 2019-04-15
Fix traefik config for webhook service

ableuler:
See `Regex Syntax` described here: https://docs.traefik.io/v2.0/routing/routers/#rule


jirikuncar:
@ableuler thanks for spotting it 👀

PR #124, opened 2019-04-17
Fix infinite redirection on login

### LINKED ISSUE #123: Fix infinte redirection on login
Depending on the order in which services come up during deployment, there are cases where the UI enters an infinite redirection loop when trying to log in.

ableuler:
Both, keycloak and gitlab offer an endpoint to query the relevant URLs for their role as openId connect provider. We make sure that these enpoints are queried repetedly until the two services are actually up.
Closes #123.

### DIFF
--- a/app/auth/gitlab_auth.py
+++ b/app/auth/gitlab_auth.py
@@ -80,14 +80,16 @@ class GitlabUserToken():
 
 SCOPE = ['openid', 'api', 'read_user', 'read_repository']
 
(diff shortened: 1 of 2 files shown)


rokroskar:
Do you have a reliable way of reproducing the bug to test this PR?

ableuler:
Purging and redeploying the entire platform has reliably led to the problem. This PR fixes the issue for me.

PR #131, opened 2019-04-23
Chart cleanup

ableuler:
Some light cleanup of the gateway chart. Some more should follow once the client applications that we create (especially inside keycloak) have been consolidated.


helm-chart/renku-gateway/values.yaml
```diff
@@ -68,7 +57,7 @@ image:
   tag: 'v2.0'
   pullPolicy: IfNotPresent
 
-  ## Define image for auth middleware
+  ## Define the image for the auth middleware
```
rokroskar: I don't even have to look at the blame to know who wrote this comment! amirite @jirikuncar ?? :)

PR #136, opened 2019-05-02
Base-64 encode redis keys

ableuler:
Federated identities can come with characters that are not supported as redis keys. Base 64 encoding those keys solves that issue.

### DIFF
--- a/app/auth/web.py
+++ b/app/auth/web.py
@@ -170,7 +170,11 @@ def get_valid_token(headers):
 
 
 def get_key_for_user(token, name):
-    return "cache_{}_{}".format(token.get('sub'), name)
+    """ Create a base-64 encoded key for the redis store """
+    from base64 import b64encode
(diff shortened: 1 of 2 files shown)


app/auth/web.py
```diff
@@ -170,7 +170,11 @@ def get_valid_token(headers):
 
 
 def get_key_for_user(token, name):
-    return "cache_{}_{}".format(token.get('sub'), name)
+    """ Create a base-64 encoded key for the redis store """
+    from base64 import b64encode
+    key = 'cache_{}_{}'.format(token.get('sub'), name)
+    return b64encode(key.encode()).decode('utf-8')
+
```
fgeorgatos: 1 question on this: is it backward-compatible?

ableuler: For an existing deployment all active sessions will be killed when upgrading the GW to this image, the best will be to also redeploy redis when changing to this image.

PR #140, opened 2019-05-08
Configure a development mode for traefik

ableuler:
Activating the development mode should not be done in production as it causes traefik to
- be more verbose
- be more generous on AllowedHosts, SSL and STS

### DIFF
--- a/helm-chart/renku-gateway/templates/configmap.yaml
+++ b/helm-chart/renku-gateway/templates/configmap.yaml
@@ -10,6 +10,15 @@ metadata:
     heritage: {{ .Release.Service }}
 data:
   traefik.toml: |
+    {{ if .Values.development }}
+    [Global]
+      debug = true
+    [log]
+      level = "debug"
+    {{ else }}
+    [log]
+      level = "error"
+    {{ end }}
     [api]
       dashboard = true
     [providers]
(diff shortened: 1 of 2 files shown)


rokroskar:
What are the benefits? Seems to me that the logging level should be separate from the security aspects.

ableuler:
> What are the benefits? Seems to me that the logging level should be separate from the security aspects.

One direct benefit of the reduced security is that it allows for setups where you mix http and https services (minikube and gitlab.com). Having one switch for a variety of settings suited for development is not at all unusual. The `NODE_ENV=development` setting for example has a variety of consequences throughout the entire application stack (https://dzone.com/articles/what-you-should-know-about-node-env).

rokroskar:
> Having one switch for a variety of settings suited for development is not at all unusual.

oh I agree - I meant only that turning on extra logging might be useful also in production.

ableuler:
> I meant only that turning on extra logging might be useful also in production.
For those cases I thought that quickly editing the configmap directly would be ok for now, but if you prefer I can add a logLevel to the chart that - when set - will overwrite the defaults set by the `production/development` switch.

rokroskar:
You mean the configmap for the deployment? I guess that would make the pod restart with the updated toml file. Doesn't the traefik chart provide a logging option?

ableuler:
Traefik is not a requirement of the renku-gateway chart. As far as I know there isn't a chart available yet for traefik 2.0. And yes, I meant exactly that: Editing the `renku-gateway` configmap would result in a changed toml, although I think you have to kill the old pod manually for this.

rokroskar:
Oh I see. In that case 👍 

> although I think you have to kill the old pod manually for this

if you change the deployment, the pod will be recreated by k8s.

ableuler:
> if you change the deployment, the pod will be recreated by k8s.

yes, but I'm editing the configmap, not the deployment. This won't recreate the pod automatically, I just checked.

I also update #113 to enable setting a log level through the chart.

PR #142, opened 2019-05-08
WIP: Fix traefik routing

ableuler:
- Remove hardcoded `/gitlab` and `/jupyterhub` path service prefixes.
- Remove all defaults for Gitlab and Jupyterhub origins on application and chart level. Both
must be set explicitly.

Note that traefik strips the path from the server url of its service definitions. The service
prefix has to be added through a path AddPrefix middleware. Therefore, we need origin and
servicePrefix as separate chart values.

---
**Do not merge** before corresponding changes have been made to the main Renku chart.


rokroskar:
If you don't want a PR to be merged, make it a draft pr.

helm-chart/renku-gateway/templates/configmap.yaml
```diff
@@ -79,10 +79,10 @@ data:
       [http.middlewares.development.headers]
         isDevelopment = true
       [http.middlewares.gitlab.AddPrefix]
-        prefix = "/gitlab/api/v4"
+        prefix = "{{ .Values.global.gitlab.servicePrefix }}/api/v4"
       [http.middlewares.jupyterhub.ReplacePathRegex]
         regex = "^/jupyterhub/(.*)"
-        replacement = "/jupyterhub/hub/api/$1"
+        replacement = "{{ .Values.global.jupyterhub.servicePrefix }}/api/$1"
```
rokroskar: I think `/hub` is there by default

ableuler: > If you don't want a PR to be merged, make it a draft pr.

I would have, but you can't go back once you've pressed the wrong button...

ableuler:
> If you don't want a PR to be merged, make it a draft pr.
I would have, but once you've pressed the wrong button there's no going back...

rokroskar:
I know, that was meant for future reference 😁

The wip in the title also works

ableuler:
outdated, closing!

PR #148, opened 2019-05-28
feat: make Keycloak realm configurable

### LINKED ISSUE #147: Make realm name configurable
Currently, the gateway assumes the Keycloak realm to be called `Renku`. This should be configurable.

ableuler:
So far the use of a Keycloak realm called 'Renku' was hardcoded. This can now be set through the charts, the default is set at the application level to 'Renku'.

Closes #147

### DIFF
--- a/helm-chart/renku-gateway/templates/deployment.yaml
+++ b/helm-chart/renku-gateway/templates/deployment.yaml
@@ -125,6 +125,10 @@ spec:
               value: {{ .Values.jupyterhub.clientId | quote }}
             - name: KEYCLOAK_URL
               value: {{ .Values.keycloakUrl | default (printf "%s://%s" (include "gateway.protocol" .) .Values.global.renku.domain) | quote }}
+              {{ if .Values.global.keycloak.realm }}
+            - name: KEYCLOAK_REALM
+              value: {{ .Values.global.keycloak.realm | quote }}
+              {{ end }}
             - name: GATEWAY_SERVICE_PREFIX
               value: {{ .Values.gatewayServicePrefix | default "/api/" | quote }}
             - name: GATEWAY_REDIS_HOST
--- a/helm-chart/renku-gateway/values.yaml
+++ b/helm-chart/renku-gateway/values.yaml
@@ -23,6 +23,14 @@ global:
     ## set by parent chart.
     domain: example.local
 
+    ## Set the Keycloak Realm name used by Renku here - the default 
+    ## value is "Renku" and is set on the application level. You may 
+    ## override this here if you are using an external Keycloak instance 
+    ## and want to use an existing realm. If Keycloak is deployed as a 
+    ## part of Renku DO NOT change this value.
+    # keycloak:
+    #   realm: Renku
+
 replicaCount: 1
 
 ## Set to true to enable the developement mode. This has negative security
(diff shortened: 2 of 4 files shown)


helm-chart/renku-gateway/values.yaml
```diff
@@ -23,6 +23,11 @@ global:
     ## set by parent chart.
     domain: example.local
 
+    ## The name of the Keycloak Realm to be used is set to 'Renku'
+    ## as a default on the application level. It can be overwritten through a chart.
```
rokroskar:
```suggestion
    ## as a default on the application level. It can be overwritten through the chart template.
```

helm-chart/renku-gateway/templates/deployment.yaml
```diff
@@ -125,6 +125,10 @@ spec:
               value: {{ .Values.jupyterhub.clientId | quote }}
             - name: KEYCLOAK_URL
               value: {{ .Values.keycloakUrl | default (printf "%s://%s" (include "gateway.protocol" .) .Values.global.renku.domain) | quote }}
+              {{ if .Values.global.keycloak.realm }}
```
rokroskar: If the realm is set to something other than `Renku`, should we not change the name in the realm we inject into keycloak?

ableuler: Yes, I agree. For me the post-install job for keyclaok should
- create a realm with the chosen name if it doesn't exist already
- add the necessary clients to the realm of the chosen name if they do not exist already

This should be implemented when removing the realm template and using the API instead. For now this is just a small change on the gateway level which does not break need changes / break the renku chart.

rokroskar: but using a different name for the realm will break the deployment.

ableuler: Sure, if that realm doesn't exist. But why would anybody do that as long as this option is not described in the main renku chart? For now and just by itself that PR doesn't change much, except allowing for some later changes to the Renku chart and enabling a use case where you have a previously existing realm and would like to use it instead of creating a new one.

rokroskar: Sure -- but then please document the value clearly and with a strong warning that if using the renku-deployed keycloak it should not be changed.

ableuler: 👍 done!

helm-chart/renku-gateway/values.yaml
```diff
@@ -23,6 +23,18 @@ global:
     ## set by parent chart.
     domain: example.local
 
+    ## The name of the Keycloak Realm to be used is set to 'Renku'
+<<<<<<< Updated upstream
```
rokroskar: there seems to be some merge issue here?

helm-chart/renku-gateway/values.yaml
```diff
@@ -23,6 +23,14 @@ global:
     ## set by parent chart.
     domain: example.local
 
+    ## The name of the Keycloak Realm to be used is set to 'Renku'
```
rokroskar: I find this wording a bit confusing -- how about:

"Set the Keycloak Realm name used by Renku here - the default value is "Renku". You may change this if you are using an external Keycloak instance and want to use an existing realm. If Keycloak is deployed as a part of Renku DO NOT change this value."

ableuler:
please re-approve, there was a typo still.

PR #151, opened 2019-06-10
Improve traefik settings

ableuler:
- Restructure traefik configuration
- Introduce rate limits, make them configurable through chart values. The default chart values for the rate limits are 20 req per 10s (burstable up to 100 req / 10s) in general and 10 req per 3s (burstable up to 10 req / 10s) for the notebooks service. Violation of those limits will result in a `429 - too many requests` response.

---
The PR is currently deployed under https://andreas.dev.renku.ch

### DIFF
--- a/helm-chart/renku-gateway/values.yaml
+++ b/helm-chart/renku-gateway/values.yaml
@@ -37,6 +37,26 @@ replicaCount: 1
 ## implications and should never be done in a production setting.
 development: false
 
+## To protect the backend services from an excessive amount of API calls
+## issued by one client, one can enforce rate limits here. The limits apply
+## per UI client session (identified by the cookies). For an explanation of
+## the different values check out the rate limiting documentation of traefik
+## v2.0.
+rateLimits:
+  ## General rate limit, applies to all /api calls combined.
+  general:
+    extractorfunc: request.header.cookies
+    period: 10s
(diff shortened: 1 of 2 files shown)


helm-chart/renku-gateway/values.yaml
```diff
@@ -37,6 +37,26 @@ replicaCount: 1
...
+  general:
+    extractorfunc: request.header.cookies
+    period: 10s
+    average: 20
+    burst: 100
+  ## Special rate limit for the notebooks service (/api/notebooks)
+  notebooks:
+    extractorfunc: request.header.cookies
+    period: 10s
+    average: 3
```
lorenzo-cavazzi: With `3` here, sometimes I get errors from the server endpoint since it may be invoked every 3 seconds for a while. 

Screenshot from 2019-06-12 13-23-47

4 should be enough, I would go for 5 to be safe -- it's enough to avoid getting errors with a single tab open, but not enough for 2 tabs.

ableuler: Ok, I will move to 5 here 👍

ableuler:
Changed to 5 requests per second for the notebooks server.
@lorenzo-cavazzi do you want to create a corresponding issue in the UI repo for the user alert?

PR #155, opened 2019-07-01
fix(charts): re-adjust notebooks rate limits

lorenzo-cavazzi:
We have recently raised the rate limits for requests to the notebooks service (#154) but it is occasionally reached when starting a new notebook with SwissDataScienceCenter/renku-ui#472 . Considering that SwissDataScienceCenter/renku-notebooks#177 has been merged, we can safely relax it a little bit more.


lorenzo-cavazzi:
yes, it probably makes sense since they are not so different anymore

PR #157, opened 2019-07-04
Jena basic auth

rokroskar:
Use basic auth with Jena.


rokroskar:
Need a review from @SwissDataScienceCenter/renku-gateway-maintainers ;)

PR #161, opened 2019-07-16
fix ontology definition

rokroskar:
Using `dcterms` is no longer necessary and we are using `schema.org` instead - therefore the graph query needs to be modified.


rokroskar:
Right, I suppose there should be an `OR` in the query...

lorenzo-cavazzi:
ok, that should be fixed in `renku-graph` since the graph rebuilt is done there, right?

rokroskar:
> ok, that should be fixed in renku-graph since the graph rebuilt is done there, right?

you mean the renku-python version? Yes - but I think for now we can just check for both predicates.

rokroskar:
can you try again? I just pushed a change that should make it work with the old ontology as well.

rokroskar:
Presumably for a short time. This will happen frequently over the next weeks as we converge on a graph that we like unfortunately...

PR #166, opened 2019-08-20
chore: bump version and changelog

rokroskar:


helm-chart/renku-gateway/Chart.yaml
```diff
@@ -3,4 +3,4 @@ appVersion: '2.0'
 description: A Helm chart for the Renku gateway
 name: renku-gateway
 icon: https://github.com/SwissDataScienceCenter/renku-sphinx-theme/raw/master/renku_sphinx_theme/static/favicon.png
-version: 0.4.2
+version: 0.5.1
```
lorenzo-cavazzi: shouldn't this be `0.5.0` for now?

rokroskar: This is the version of the next release, actually -- all the automatically-generated tags will have e.g. `0.5.1-abcd1234` which is a development version of 0.5.1.

rokroskar: I tagged the current one with 0.5.0 because of the API change

PR #181, opened 2019-11-27
feat: add core service routing

### LINKED ISSUE #180: Relay calls to core-service through the gateway
The new Renku repository service will need to be exposed through the gateway. My suggestion would be to expose it as `/api/repository/...` which will result in API endpoints such as 
```
https://renkulab.io/api/repository/cache/project-clone
```
To serve such requests, the repository service will need access to the gitlab oauth access token and some user information:
```
Renku-user-id: 9af6abce-bd63-48e4-a0c7-233a8c524e48
Renku-user-fullname: Andreas Bleuler
Renku-user-email: andi@example.com
Authorization: Bearer db6d049ec8f3c225c248fe6f3caddbd47c948edb22e6b83b37cc08d79b264ca1
```
All these will be added to dedicated header fields by the gateway before forwarding the request.

The repository URL (containing gitlab namespace and project name) will be provided by the client as part of the request body:
```
{
  "project_url": "https://dev.renku.ch/gitlab/andreas/coffee-machine"
}
```

rokroskar:
- [x] add traefik routes for the core service
- [x] add the required renku-user headers (see #180)

closes #180

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -86,12 +86,12 @@ async def auth():
 
     from .auth.gitlab_auth import GitlabUserToken
     from .auth.jupyterhub_auth import JupyterhubUserToken
-    from .auth.renku_auth import RenkuAuthHeaders
+    from .auth.renku_auth import RenkuCoreAuthHeaders
 
     auths = {
         'gitlab': GitlabUserToken,
         'jupyterhub': JupyterhubUserToken,
-        'renku': RenkuAuthHeaders,
+        'renku': RenkuCoreAuthHeaders,
     }
 
     auth = auths[request.args.get('auth')]()
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -86,10 +86,12 @@ async def auth():
 
     from .auth.gitlab_auth import GitlabUserToken
     from .auth.jupyterhub_auth import JupyterhubUserToken
+    from .auth.core_auth import CoreAuthHeaders
 
     auths = {
         'gitlab': GitlabUserToken,
         'jupyterhub': JupyterhubUserToken,
+        'core': CoreAuthHeaders,
     }
 
     auth = auths[request.args.get('auth')]()
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -86,12 +86,12 @@ async def auth():
 
     from .auth.gitlab_auth import GitlabUserToken
     from .auth.jupyterhub_auth import JupyterhubUserToken
-    from .auth.core_auth import CoreAuthHeaders
+    from .auth.renku_auth import RenkuAuthHeaders
 
     auths = {
         'gitlab': GitlabUserToken,
         'jupyterhub': JupyterhubUserToken,
-        'core': CoreAuthHeaders,
+        'renku': RenkuAuthHeaders,
     }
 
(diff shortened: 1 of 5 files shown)


rokroskar:
I think the main thing to decide is what to call the routes. Right now it's `/core` but `/renku` has been suggested.

ableuler:
> I think the main thing to decide is what to call the routes. Right now it's `/core` but `/renku` has been suggested.

Would it be possible to 
1. Make the distinction between RPC and REST a bit clearer in the URL. For me this would mean that RPC endpoints ideally remain "flat".
2. Choose the correct backend service based on the resource in the case of REST (we're already doing this) or the desired action in the case of RPC. Very likely all RPC endpoints would be handled by the `renku-core` service.  

So instead of 
```
https://renkulab.io/api/core/datasets/create
```
we'd have
```
https://renkulab.io/api/createDataset.
```

rokroskar:
But then we have to create a route per command? Wouldn't some top-level grouping be useful?

ableuler:
> But then we have to create a route per command? Wouldn't some top-level grouping be useful?

Maybe we can apply some general rule? Something like this:
```
/camelCase => RPC => forward to renku-core
```
But in general you are definitely right: A clear indication about the right backend service already in the URL will make the rules in the gateway much easier. Avoiding this on the other hand would make for nicer URLs and gives us more flexibility to change things internally without breaking the API. Let's say we suddenly want to serve `/projects` from the graph service instead of gitlab. Currently, we can do this without modifying the UI.

lorenzo-cavazzi:
+1 for grouping all the rpc calls (`/api/core/` or `/api/renku/`), but without further mixing in the "rest style" -- so I would prefer `/createDataset` over `/datasets/create`.

The final result would be something like this `https://renkulab.io/api/renku/{function}`

```
https://renkulab.io/api/renku/createDataset
```

rokroskar:
I like this idea @lorenzo-cavazzi - ideally the routes would then also follow the same convention in the service itself. @jsam?

ableuler:
My only small concern with the suggestion of @lorenzo-cavazzi is that if I see `/api/renku`, I would expect to find all renku-specific endpoints there (including graph stuff, notebook servers, etc) and not just rpc endpoints which correspond Renku cli commands.

jsam:
I find having camel case notation mixed in the URL and as method name very un-natural. If we don't like method reference name in the URL it makes more sense then to go for full jsonrpc 2.0 compliance instead and drop the OpenAPI spec.

rokroskar:
@jsam and I discussed this a bit further - basically at the moment there doesn't seem to be a nice way to automatically generate docs from the code for pure RPC. I propose that we leave the API as is for now, but call it `/renku` instead of `/core`.

Panaetius:
One way I like for urls in this scenario is: `https://renkulab.io/api/renku/dataset.create`

It's still a flat list, the dot has no significance technically, but it helps delineate the methods and it reads like class methods, so it's easy to understand. And doesn't depend on case

jsam:
@Panaetius Done. Service will work with dot-based naming schema for methods. 

I think we can conclude this discussion and merge this.

PR #185, opened 2020-02-04
build: automatically update renku-gateway dependency in renku

ableuler:
Note that the travis CLI did some auto-modifications to .travis-ci.yaml.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,54 +1,45 @@
 notifications:
   email: false
-
 sudo: required
-
 services:
-  - docker
-
+- docker
 language: python
-
 python:
-  - 3.7-dev
-
+- 3.7-dev
 env:
   global:
-    - DOCKER_LABEL=$TRAVIS_BRANCH
-    - DOCKER_PUSH_LATEST=true
-    - DOCKER_REPOSITORY=renku/
-    - HELM_URL=https://storage.googleapis.com/kubernetes-helm
-    - HELM_TGZ=helm-v2.9.1-linux-amd64.tar.gz
-
+  - DOCKER_LABEL=$TRAVIS_BRANCH
+  - DOCKER_PUSH_LATEST=true
+  - DOCKER_REPOSITORY=renku/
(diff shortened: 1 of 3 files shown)


rokroskar:
Instead of pushing directly to master and having to give the bot admin rights to the repo, can we instead push to a branch, say `/renku/auto-update/renku-gateway-<commit-sha>`?

cc @jachro: I think we should change this in `renku-graph` as well. I will make an action in the `renku` repo to automatically open PRs for these branches so there won't really be any need for manual intervention.

ableuler:
> instead push to a branch, say `/renku/auto-update/renku-gateway-<commit-sha>`?

We agreed on using `/renku/auto-update/renku-gateway-<chart-version>` to distinguish tagged releases from regular ones.

.travis.yml
```diff
@@ -1,54 +1,51 @@
...
 services:
-  - docker
-
+- docker
 language: python
-
 python:
-  - 3.7-dev
-
+- 3.7-dev
```
rokroskar: I know you didn't just change this but why are we using a dev version?

.travis/travis-functions.sh
```diff
@@ -0,0 +1,49 @@
...
+function updateVersionInRenku() {
+  echo "Updating renku-gateway version to $CHART_VERSION in Renku's requirements.yaml"
+
+  # cloning the renku repo
+  cd || exit
+  git clone https://github.com/SwissDataScienceCenter/renku.git
+  cd renku || exit
+
+  # changing the version
+  sed -i "/- name: renku-gateway/{n;n;n;s/.*/  version: $CHART_VERSION/}" charts/renku/requirements.yaml
```
rokroskar: in the future we should use `yq` to do this in a more human-readable way maybe

.travis.yml
ableuler: I don't remember, likely that's not necessary anymore.

.travis/travis-functions.sh
ableuler: Agreed.

PR #202, opened 2020-05-06
fix: change logout method to POST

lorenzo-cavazzi:
GitLab 12.7.0 prevented triggering GitLab logout from outside GitLab, 12.9.0 brought it back. However, a POST request is necessary now. 

My(@ableuler) suggestion is actually to keep this PR open for 2-3 months until more people will be ready to migrate their GitLab instance to `>=12.9.0`.

### DIFF
--- a/app/auth/gitlab_auth.py
+++ b/app/auth/gitlab_auth.py
@@ -29,7 +29,16 @@ from oic.oic import Client
 from oic.oic.message import AuthorizationResponse, RegistrationResponse
 from oic.utils.authn.client import CLIENT_AUTHN_METHOD
 from oic.utils.keyio import KeyJar
-from quart import Blueprint, Response, current_app, redirect, request, session, url_for
+from quart import (
+    Blueprint,
+    Response,
+    current_app,
+    redirect,
+    render_template,
+    request,
+    session,
+    url_for,
+)
 
 from .web import JWT_ALGORITHM, get_key_for_user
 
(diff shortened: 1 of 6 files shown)


app/auth/gitlab_auth.py
```diff
@@ -238,5 +239,8 @@ def get_gitlab_refresh_token(access_token):
 @blueprint.route('/logout')
 def logout():
     logout_url = current_app.config['GITLAB_URL'] + '/users/sign_out'
-    response = current_app.make_response(redirect(logout_url))
-    return response
+    return render_template(
```
rokroskar, line 242: instead of making this breaking change, could we have a configuration variable that allows us to satisfy gitlab pre and post `12.9`?

ableuler: I generally like the idea, but it's more annoying than it seems. Currently, the gateway is not aware of the gitlab version. The gitlab API has a `/version` endpoint, but only available for logged-in users, so it can't just be queried at startup. Doing this in a clean way would therefore need non-negligible chart refactoring. I agree that we will eventually have to support multiple versions of the same dependency, as long as we can avoid this work by upgrading RenkuLab and GitLab in the SDSC deployments I would prefer avoiding this a little longer.

rokroskar: why does it need to be automatic? It can be a chart value that whoever is deploying the platform can set.

ableuler: done as discussed

PR #204, opened 2020-05-11
chore: update dependencies

ableuler:


Pipfile
```diff
@@ -8,15 +8,15 @@ werkzeug = "*"
...
-quart = "*"
-quart-cors = "*"
 pyjwt = "*"
 redis = "*"
 flask-kvsession = "*"
 aiohttp = "*"
 sparqlwrapper = "*"
 responses = "*"
 aioresponses = "*"
+quart = "<0.11.0"
```
lorenzo-cavazzi, line 18: Wouldn't it be safer to use `==` instead of `<`? Otherwise, another dependency in the future may force installing a lower version (ok, that's unlikely... but still better to be safe :slightly_smiling_face:  )

ableuler: I want it to be able to upgrade to a higher patch version, so from `0.10.z` to `0.10.z+1`. It seems that `0.11` breaks it.

Issue #192, opened 2020-03-27
Refresh tokens in the gateway whenever possible


ableuler:
The gateway should always and independently try to refresh its token without involving the UI. If the gateway can not refresh any token, it should respond with: 

```json
{
   "statusCode": "401,Unauthorized",
    "error":"invalid_token",
    "message":"The access token expired",
}
```

This will trigger a full logout and re-login procedure.

PR #212, opened 2020-05-11
Use black formatting

ableuler:


rokroskar:
you could also set up a pre-commit hook: https://github.com/SwissDataScienceCenter/renku-notebooks/blob/master/.pre-commit-config.yaml#L2-L6

ableuler:
> Nice! ⚫
> 
> Should we also add [pytest-black](https://pypi.org/project/pytest-black/) and enforce black formatting in the tests with `pytest --black`? Otherwise, the formating wouldn't be mandatory for new PRs.

Thanks @lorenzo-cavazzi. Would you mind adding this to the present PR? 🙏 
I mostly wanted to have a somewhat cleaner state before tackling https://github.com/SwissDataScienceCenter/renku-gateway/issues/192 and now I'm neck-deep inside that.

ableuler:
> you could also set up a pre-commit hook: https://github.com/SwissDataScienceCenter/renku-notebooks/blob/master/.pre-commit-config.yaml#L2-L6

I was wondering about that recently: This requires you to have a system-wide install of black, right? (not that this would be a deal-breaker...)

rokroskar:
@ableuler I run `chartpress` with `pipenv` here: https://github.com/SwissDataScienceCenter/renku-notebooks/blob/master/.pre-commit-config.yaml#L17 - I'm not sure how it would work with `black`.

lorenzo-cavazzi:
Why is `pytest` now going up to `../../../`? :thinking: 
Has anyone already seen this?

lorenzo-cavazzi:
sorry for forcing, I did (too) many attempts at fixing the failing tests. I didn't manage to limit the scope of `pytest` properly, I can't understand where the issue comes from

ableuler:
@lorenzo-cavazzi your changes seemed all fine to me, there was just a problem with some library not getting installed. I recreated the lockfile which fixed the issue and squashed everything into two commits.

PR #222, opened 2020-06-02
Revert back from quart to flask

ableuler:
This PR reverts back from the asynchronous quart framework to the more standard, easier to maintain flask framework.


rokroskar:
I'm also wondering whether it's worth ripping quart out now? It was quite some work to get it working correctly iirc.

ableuler:
Oh, ok - I didn't imagine that this would face any opposition, so I was short on motivating this PR. Let me give you some reasons for doing this (and for doing it now):

- Flask is more widely used and therefore has better documentation and tooling than quart. Quart almost always works like flask, but when it doesn't, then it get's really annoying. Quart also forces you to use less common (since asynchronous) libraries for other things such as doing http requests, hence testing, etc... 

- The reason for using quart was the proxy part of the gateway. With that part being handled by traefik now, the most IO-bound actions happening in `gateway-auth` are the redis calls. The standard `redis-py` library is not asynchronous, which kind of makes the use ob quart a bit absurd. There is an `aioredis` library that we could use, but that would be more refactoring and another non-standard way of doing things.

- Developing in flask is more pleasant than in quart. When messing up in quart, you for example simply get a 500 without any traceback if you for example forget to await an async function call. That was actually what annoyed me to the point where I invested the 30 minutes to replace quart back with flask. 

- Currently we can't upgrade the quart version. I didn't look yet at what exactly breaks when upgrading, might be trivial or hard to fix.

- As @rokroskar said it took some time to get this to work. But for the same reasons that made it time consuming to put it there in the first place it will remain a burden to maintain.

- @lorenzo-cavazzi I would argue it's even the other way round. If we want to implement websockets, I'd rather do this using [flask-socketio](https://flask-socketio.readthedocs.io/en/latest/) which again seems more used and therefore has more documentation and examples than quart.

In conclusion I'm convinced that while we did have a reason for using quart and it was justified putting in that work back then, we're making our life much easier by going back to flask.

Issue #252, opened 2020-07-24
Handle unicode characters in headers sent to core service


rokroskar:
Currently any user with a non-latin-1-encodable character in their full name will receive a 500 response from the gateway when trying to use any of core service APIs. The reason is that headers need to be latin-1 encodable. We should handle this appropriately in the gateway in one of two ways:

a) simply base64-encoding the headers 
b) putting the JWT itself (where this information comes from) into the header

Since the bug currently prevents users with non-latin-1 encodable characters, we should patch it asap with the simplest solution (a) and discuss whether it's worthwhile implementing solution (b).

ableuler:
@rokroskar @lorenzo-cavazzi 
This has been closed by #253, right?

rokroskar:
well partially - we haven't implemented the full solution (using the JWT directly) that's why it was kept open

ableuler:
As discussed on [this issue](https://github.com/SwissDataScienceCenter/renku-python/issues/1414) the gateway should move the entire ID token of the user in a custom header.

PR #253, opened 2020-07-24
fix: avoid crash on fullname non-latin-1 chars

lorenzo-cavazzi:
Very minimalistic fix for #252
The `Renku-user-fullname` gets encoded to base64 and sent with the `b` prexif (E.G. `"b'TG9yZW7FoW8gQ2F2YXp6aQ==\\n'"`). This (apparenty) results in automatic decoding from the receiving party.

We should switch to a more robust solution, like the other one suggested in #252

### DIFF
--- a/app/auth/renku_auth.py
+++ b/app/auth/renku_auth.py
@@ -17,6 +17,7 @@
 # limitations under the License.
 """Add the headers for the Renku core service."""
 
+import base64
 import re
 
 from flask import current_app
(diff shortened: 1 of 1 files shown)


app/auth/renku_auth.py
```diff
@@ -46,7 +47,9 @@ def process(self, request, headers):
             access_token_dict = decode_keycloak_jwt(access_token.encode())
             headers["Renku-user-id"] = access_token_dict["sub"]
             headers["Renku-user-email"] = access_token_dict["email"]
-            headers["Renku-user-fullname"] = access_token_dict["name"]
+            headers["Renku-user-fullname"] = str(
```
rokroskar, line 50: why do you need the `str()`?

rokroskar:
This has been tested against https://github.com/SwissDataScienceCenter/renku-python/pull/1407 and it seems to work fine.

PR #254, opened 2020-07-24
build: pin ptvsd

lorenzo-cavazzi:
For some reason, I can't debug with the pre-release version of `ptvsd`. The last official release is `4.3.2` and it's the last one working. Am I the only one experiencing this problem?
P.S. I opened a draft PR instead of an issue to better describe the possible solution :smile:


lorenzo-cavazzi:
This doesn't seem to be a problem aynmore

PR #280, opened 2020-09-01
fix(charts): fix a bug in the direct routing to gitlab

ableuler:
The recent changes allowing to query non-api gitlab endpoints (https://github.com/SwissDataScienceCenter/renku-gateway/pull/251) introduced an error which manifests when gitlab is not running under `/gitlab`. This PR fixes it.

### DIFF
--- a/helm-chart/renku-gateway/templates/configmap.yaml
+++ b/helm-chart/renku-gateway/templates/configmap.yaml
@@ -96,7 +96,14 @@ data:
         [http.routers.direct]
           # This is to access undocumented APIs in GitLab
           entryPoints = ["http"]
-          Middlewares = ["direct", "gitlabOnly"]
+          Middlewares = [
+            "direct"
+            {{- if .Values.global.gitlab.urlPrefix -}}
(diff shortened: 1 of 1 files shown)


helm-chart/renku-gateway/templates/configmap.yaml
```diff
@@ -96,7 +96,14 @@ data:
         [http.routers.direct]
           # This is to access undocumented APIs in GitLab
           entryPoints = ["http"]
-          Middlewares = ["direct", "gitlabOnly"]
+          Middlewares = [
+            "direct"
+            {{- if .Values.global.gitlab.urlPrefix -}}
```
rokroskar, line 101: can't you do this with one if statement?

rokroskar (approved):
looks good - the double if statement looks unnecessary to me, but if it's the only way then it's fine

ableuler:
> looks good - the double if statement looks unnecessary to me, but if it's the only way then it's fine

I simply wasn't sure if the go template engine respects the order of if statements in a long chain of ifs. Do you know?

PR #285, opened 2020-09-03
chore: move to github native dependabot

ableuler:
~Together with merging this PR, the `dependabot preview` app should be deactivated for this repo~, see https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/.

We can merge this PR and see if it will result in dependency update PRs to the dedicated `dependabot-updates` branch. The idea would be to have auto-updates to a specific persistent branch and merge this branch into master occasionally.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    target-branch: "dependabot-updates"


ableuler:
I guess this configuration has to be in master in order to have any effect?

rokroskar:
most likely

.github/dependabot.yaml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
rokroskar, line 6: that's a pretty high frequency - the default setting now is once a week

.github/dependabot.yaml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
```
rokroskar, line 3: isn't it technically pipenv?

.github/dependabot.yaml
ableuler: My idea was to merge those automatically into the `dependabot-updates` all the time and then merge `dependabot-updates` into master once a week. This would give us one PR to master per week instead of once a week several PRs to master. Not sure if that will work, but that's what I wanted to try :)

.github/dependabot.yaml
ableuler: Didn't see that option [here](https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#package-ecosystem), was hoping that it will work with pipenv too

.github/dependabot.yaml
```diff
@@ -0,0 +1,7 @@
...
+    directory: "/"
+    schedule:
+      interval: "daily"
+    target-branch: "dependabot-updates"
```
rokroskar, line 7: I believe `target-branch` is the branch to check for updates, not the branch to push the updates to.

rokroskar: i.e. the default is `master`

ableuler: The way I understand the docs it's the branch which is looked at for the pip- or lockfile AND the target branch of the created PR: https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates#target-branch

rokroskar: ok I see what you mean - lets try it

rokroskar (approved):

PR #299, opened 2020-09-29
feat(auth): pass on KC id token to core service

### LINKED ISSUE #252: Handle unicode characters in headers sent to core service

ableuler:
This PR fixes obtaining the ID token from keycloak during login by making sure the defined scope is picked up by oauthlib (use `scope` instead of `scopes`) and ads the id token to a request which is forwarded to the core service in the custom header field `"Renku-User"` .

The original header fields are left there for backwards compatibility until the core service has adapted.

Closes #252

### DIFF
--- a/app/auth/gitlab_auth.py
+++ b/app/auth/gitlab_auth.py
@@ -82,7 +82,7 @@ class GitlabUserToken:
         return headers
 
 
-SCOPES = ["openid", "api", "read_user", "read_repository"]
+SCOPE = ["openid", "api", "read_user", "read_repository"]
 
 
 @blueprint.route("/login")
@@ -96,7 +96,7 @@ def login():
         provider_app,
         urljoin(current_app.config["HOST_NAME"], url_for("gitlab_auth.token")),
         GL_SUFFIX,
-        SCOPES,
+        SCOPE,
     )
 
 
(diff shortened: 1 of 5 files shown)


rokroskar (approved):
changes LGTM but should be verified with https://github.com/SwissDataScienceCenter/renku-python/pull/1520

rokroskar:
@jsam could you test this together with https://github.com/SwissDataScienceCenter/renku-python/pull/1520?

jsam:
Yes. As soon as I get to do the rebase there, I'll test the whole thing

rokroskar:
What are we waiting for with this PR? Can we get this deployed together with SwissDataScienceCenter/renku-python#1520? Or has this already been tested? If yes, then we should get this reviewed ASAP and merged.

rokroskar (approved):
LGTM - the only thing I would maybe add is a `# TODO` comment where the "old" headers can/should eventually be removed.

ableuler:
> # TODO comment where the "old" headers can/should eventually be removed.

done!

rokroskar (approved):

Issue #267, opened 2020-08-18
Enable partial logout


ableuler:
The gateway should provide a way for the UI to trigger a logout of gitlab and jupyterhub but not keycloak. Such a partial logout - followed by a login circle (which will typically not require a user password) - will fix any inconsistent login states where a user might be logged into Renku as user A but into GitLab as user B without noticing.

See also https://github.com/SwissDataScienceCenter/renku-ui/issues/879.

lorenzo-cavazzi:
Not sure if we still want to try to address this. The reference UI issue has been solved differently, and I can't think about a case where the UI will ask for a partial logout. If that should always happen, I'd rather integrate that step permanently before a login instead of adding a parameter to pilot that behavior from the client (I don't think it's needed though... Maybe for supporting external GitLab instances?).

ableuler:
I agree that we should not trigger the logout from GitLab for gitlab instances that do not come as a part of RenkuLab. But that's actually quite a different issue. Since the original goal of this issue has been solved in another way, I agree, let's close this issue.

PR #306, opened 2020-10-06
fix: adjust the time buffer in the token expiration date check function

lorenzo-cavazzi:
While trying to address #267 , I found a bug in the function checking the token expiration date.
This is very likely the reason for the unnecessary re-login triggered on the UI that happens frequently when polling a specific API for a long time, and occasionally in other situations.

***How to test***
This is a bit tricky. The easiest way is lowering the access token lifespan in your dev renkulab deplyoment to 1 minute (see the image below).
Then open any of your projects and go to the environment page, so that the `/servers` API will be invoked every ~3 seconds. Without this PR, the UI will trigger a full login and refresh every minute. Running this PR (thorugh telepresence or using the gateway-auth image `lorenzocavazzitech/renku-gateway:0.9.0-0a593a8`) the token should refresh automatically in the backend without any side-effect in the UI -- and totally trasparent for the user.

Expl_token

### DIFF
--- a/app/auth/oauth_client.py
+++ b/app/auth/oauth_client.py
@@ -134,4 +134,4 @@ class RenkuWebApplicationClient(WebApplicationClient):
 
     def expires_soon(self):
         """Check if the client instance expires soon."""
-        return self._expires_at and self._expires_at < time.time() - 5
+        return self._expires_at and self._expires_at < time.time() + 5


ableuler (approved):
Thanks for finding that one! 😬

app/auth/oauth_client.py
```diff
@@ -134,4 +134,4 @@ def _fix_expiration_time(self):
 
     def expires_soon(self):
         """Check if the client instance expires soon."""
-        return self._expires_at and self._expires_at < time.time() - 5
+        return self._expires_at and self._expires_at < time.time() + 5
```
rokroskar, line 137: 🙀

ableuler: "Vorzeichen sind Glückssache", as one of my profs always used to say...

PR #308, opened 2020-10-09
docs: fix vscode debug command

lorenzo-cavazzi:
The previous command to attach the VScode debugger didn't work because flask starts multiple sub-processes.
The `--no-reload` flag is the easiest fix, at the cost of dropping the auto-refresh when the code is changed. There may be other better solutions to explore (e.g. `debugpy`,  but I didn't manage to debug properly with that).


ableuler (approved):
Would be nice to have the combination of hot reloading and vscode-debugging at some point, but this at least fixes what we currently have, so 👍 !

PR #309, opened 2020-10-09
return verbose errors on token validation exceptions

lorenzo-cavazzi:
This will
* Provide extra content to the UI when sending a 401 error related to expired/invalid tokens
* Handle an expected exception without logging the corresponding error
 ```
[2020-10-12 09:41:23,824] INFO in oauth_redis: Refreshing cache_656bba26-0863-4e12-9fbc-ec0f0e79a4e1_kc_oidc_client
[2020-10-12 09:41:23,970] WARNING in oauth_redis: Error refreshing tokens: invalid_grant Clearing client from redis.
[2020-10-12 09:41:23,975] WARNING in __init__: Error while authenticating request
Traceback (most recent call last):
File "/code/app/__init__.py", line 116, in auth
new_tokens = web.get_valid_token(headers)
File "/code/app/auth/web.py", line 97, in get_valid_token
return {"access_token": keycloak_oidc_client.access_token}
AttributeError: 'NoneType' object has no attribute 'access_token'
```

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -180,7 +180,7 @@ def auth():
             f"Error while authenticating request, unknown. Target: {auth_arg}",
             exc_info=True,
         )
-        message = {"error": "authentication", "payload": "unknown", "target": auth_arg}
+        message = {"error": "authentication", "message": "unknown", "target": auth_arg}
         return Response(json.dumps(message), status=401)
 
     return Response(json.dumps("Up and running"), headers=headers, status=200,)
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -102,7 +102,8 @@ def auth():
         "renku": RenkuCoreAuthHeaders,
     }
 
-    auth = auths[request.args.get("auth")]()
+    auth_arg = request.args.get("auth")
+    auth = auths[auth_arg]()
     headers = dict(request.headers)
 
     # Keycloak public key is not defined so error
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: app/__init__.py [hunk 1]
```diff
@@ -146,13 +147,41 @@ def auth():
...
+            exc_info=True,
+        )
+        message = {
+            "error": "authentication",
+            "message": "token expired",
+            "target": auth_arg,
+        }
+        return Response(json.dumps(message), status=401)
+    except AttributeError as error:
+        if "access_token" in str(error):
```
ableuler, line 161: How do you know that an attribute error with this message stands for a problem in refreshing the token? Is this always the case?

app/__init__.py
```diff
@@ -146,13 +147,41 @@ def auth():
...
+            current_app.logger.warning(
+                (
+                    "Error while authenticating request, can't "
+                    f"refresh access token. Target: {auth_arg}"
+                ),
+                exc_info=True,
+            )
+            message = {
+                "error": "authentication",
+                "payload": "can't refresh access token",
```
ableuler, line 171:
```suggestion
                "message": "can't refresh access token",
```
to be consistent with the above

app/__init__.py
```diff
@@ -146,13 +147,41 @@ def auth():
...
     # TODO: fix bare except
     # https://github.com/SwissDataScienceCenter/renku-gateway/issues/232
     except:  # noqa
-        current_app.logger.warning("Error while authenticating request", exc_info=True)
-        return Response(json.dumps({"error": "Error while authenticating"}), status=401)
+        current_app.logger.warning(
+            f"Error while authenticating request, unknown. Target: {auth_arg}",
+            exc_info=True,
+        )
+        message = {"error": "authentication", "payload": "unknown", "target": auth_arg}
```
ableuler, line 183:
```suggestion
        message = {"error": "authentication", "message": "unknown", "target": auth_arg}
```
to be consistent with the above

ableuler (changes requested):
Definitely helpful to have more information here.

app/__init__.py
lorenzo-cavazzi: I did quite a few trials and the error I get when it's not possible to refresh the token is always a missing `access_token`. Iirc it happens in `web.get_valid_token` when trying to create the object `{"access_token": keycloak_oidc_client.access_token}` (the error is `AttributeError("'NoneType' object has no attribute 'access_token'")`).

Not sure if this is how that function is intended to work, maybe changing the code there to return something else (or throw a specific error) could be a cleaner solution.

ableuler: Yeah, that will work because [here](https://github.com/SwissDataScienceCenter/renku-gateway/blob/a9d2657c06f0e6b847f47f5946935628b0002634/app/auth/oauth_redis.py#L97) we just return `None` if the token can not be refreshed (which ist not very nice), hence no `access_token` attribute.

I think the cleanest would be to define a custom `token_refresh_error` as part of the `oauth_client` module and throw it already [here](https://github.com/SwissDataScienceCenter/renku-gateway/blob/a9d2657c06f0e6b847f47f5946935628b0002634/app/auth/oauth_client.py#L92) when the refreshing fails. Then we could catch that error [here](https://github.com/SwissDataScienceCenter/renku-gateway/blob/a9d2657c06f0e6b847f47f5946935628b0002634/app/auth/oauth_redis.py#L91), log the warning, re-raise the error and catch it again at the same point where you catch the attribute error now. Shouldn't be too much change, but we can also put this in a separate issue if you prefer to not tackle this now.

ableuler (approved):

PR #310, opened 2020-10-09
temp: test with partial logout and token cleanup.

lorenzo-cavazzi:
re #267


lorenzo-cavazzi:
Closing the PR since this solution doesn't look good for the purpose of addressing #267 : the code here affects the login instead of the logout (the initial idea was to solve a re-login problem on the UI so it kind of made sense).

PR #311, opened 2020-10-09
Route requests to the uiservice

ciyer:
The changes necessary to allow communication with the uiservice.

### DIFF
--- a/helm-chart/renku-gateway/templates/configmap.yaml
+++ b/helm-chart/renku-gateway/templates/configmap.yaml
@@ -123,6 +123,12 @@ data:
           Rule = "PathPrefix(`{{ .Values.gatewayServicePrefix | default "/api/" }}renku`)"
           Service = "core"
 
+        [http.routers.uiserver]
+          entryPoints = ["http"]
+          Middlewares = ["auth-renku", "common"]
+          Rule = "PathPrefix(`{{ .Values.gatewayServicePrefix | default "/api/" }}ui`)"
+          Service = "uiserver"
+
       [http.middlewares]
 
         # We assume entity IDs which are URL-safe except <namespace>/<project-name>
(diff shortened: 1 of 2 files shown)


ableuler:
This looks good - the path under which the new service will be exposed and the authentication methods are still open for me but can be adapted easily once https://github.com/swissdatasciencecenter/renku-ui/issues/1043 is merged.

ableuler:
Understanding the functionality and scope of the new ui server (as a general server-side component or the UI, potentially handling sessions, persist UI state, etc) I'd suggest to expose that service at the ingress level and not behind the `/api`(ie not through the gateway).

ciyer:
> Understanding the functionality and scope of the new ui server (as a general server-side component or the UI, potentially handling sessions, persist UI state, etc) I'd suggest to expose that service at the ingress level and not behind the `/api`(ie not through the gateway).

That is an interesting proposal. What would that mean for authentication and token management?

ableuler:
I think the easiest would be for the UI service to have its own session with the frontend. We might want to even make this the the _only_ session (ie we would route all API calls from the browser through the UI service), but that would have to be discussed and could follow in a second step.

ableuler:
What's the status here? As mentioned before I would suggest to expose this UI-service directly through an ingress and not as a part of a REST api. Would it be ok to close this issue?

lorenzo-cavazzi:
@ableuler Sorry for addressing this late. If I understand correctly your suggestion, we should drop this PR and add the ingress configuration directly in the Renku repository.
I have made a PR for that. Let me know if that works SwissDataScienceCenter/renku#2025

lorenzo-cavazzi:
Not needed anymore SwissDataScienceCenter/renku#2025

PR #312, opened 2020-10-12
feat: clear session on logout

lorenzo-cavazzi:
testing session cleanup to prevent dirty states


lorenzo-cavazzi:
This doesn't seem to be needed anymore. It was initially meant to be a workaround to avoid some problems when logging out and then login in with a different account than the previous one. It doesn't seem to be a problem anymore.

Issue #191, opened 2020-03-27
Fix logout redirection for dev-like deployments


ableuler:
Currently, whenever a user logs out in one our our dev deployments, the user ends up at the login page of `dev.renku.ch` instead of `<dev-namespace>.dev.renku.ch`. Because of https://github.com/swissdatasciencecenter/renku-ui/issues/879 it would be nice to fix this if possible.

PR #320, opened 2020-10-27
fix(app): fix JupyterHub logout and logout redirection

### LINKED ISSUE #314: Logout cycle seems to skip Jupyterhub
When hitting the logout button on the RenkuLab UI, the user stays logged in with Jupterhub. This leads to very unexpected behaviour when dealing with different users in the same browser.

ableuler:
Perform all logouts in iframes on the same page. This simplifies the code, fixes the JupyterHub logout and the final logout redirection for deployments with external GitLab.

closes #314 and #191

### DIFF
--- a/app/auth/web.py
+++ b/app/auth/web.py
@@ -17,7 +17,7 @@
 # limitations under the License.
 """Web auth routes."""
 
-from urllib.parse import urljoin, urlencode
+from urllib.parse import urljoin
 
 from flask import (
     Blueprint,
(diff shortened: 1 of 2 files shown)


lorenzo-cavazzi (approved):
Very nice solution!
The logout user experience is much better without multiple redirections :+1: 

I think cleaning the reference to the session in the browser as suggested here https://github.com/SwissDataScienceCenter/renku-gateway/pull/312/commits/efcfd0c83e4e44ed17e9fd110ce8e377ae2c854e in the line `resp.delete_cookie("session")` is not needed anymore, even though it _may_ be useful when errors occur in the iframes to be sure all the references to the previous client session are gone.

PR #322, opened 2020-10-28
Dependabot updates

ableuler:


lorenzo-cavazzi:
I solved the conflicts with `master` by choosing the most recent version of the libraries and it seems that most of these updates were already there :shrug:

lorenzo-cavazzi (approved):

ableuler:
> I solved the conflicts with `master` by choosing the most recent version of the libraries and it seems that most of these updates were already there 🤷

Thanks - there was a security alert PR yesterday. Those go into master, therefore the conflict.

PR #353, opened 2021-01-05
ci: update helm url

lorenzo-cavazzi:
The test-chart action uses the helm version 2.16.9. We need to update to version >= 2.17.0 to prevent it from failing as in https://github.com/SwissDataScienceCenter/renku-ui/runs/1649868184

Reference: https://helm.sh/blog/new-location-stable-incubator-charts/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Install helm
         env:
           HELM_URL: https://storage.googleapis.com/kubernetes-helm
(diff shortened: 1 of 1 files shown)


pameladelgado (approved):
Indeed, I was going to do this. Thanks @lorenzo-cavazzi !

lorenzo-cavazzi:
Ah yeah, sorry for rushing on this. I created the commit in the UI and, since the action I use there is a copy/paste, I extended it to the other repos as well.

pameladelgado:
no worries, makes perfectly sense :slightly_smiling_face:

PR #373, opened 2021-02-10
chore: update a bunch of dependencies

ableuler:
This includes a vulnerability fix.
/deploy

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -87,16 +87,16 @@
         },
         "cryptography": {
             "hashes": [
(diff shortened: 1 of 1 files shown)


lorenzo-cavazzi (approved):
:rocket: 

I was wondering if we should pin the dependencies in `Pipfile` as well :thinking:

ableuler:
> 🚀
> 
> I was wondering if we should pin the dependencies in `Pipfile` as well 🤔

I think when using pipenv it's common practice to only do that for known incompatibilities of the code with given versions. The reproducibility of a build is already ensured through the lockfile.

PR #388, opened 2021-03-10
chore: pin cleanup action

ableuler:


lorenzo-cavazzi (approved):

rokroskar:
oh no I was just a few minutes too late... I just fixed the bug 😆

lorenzo-cavazzi:
@rokroskar Should we revert this if we are going to tag a new release?

ableuler:
> @rokroskar Should we revert this if we are going to tag a new release?

Nono, having this pinned is better anyway. We can move to the newly tagged release of that action once the "anonymous maintainer" has made sure that it really works :)

Issue #406, opened 2021-04-22
Add gitlab oauth token to notebook-service requests


ableuler:
The notebooks-service can not obtain gitlab tokens anymore once it's disconnected from JH and will need to get them from the gateway.

---

## Design suggestion
(_edit_)

It would be nice to take this issue as an example for a potential future header structure that all renku backend services would agree on. This would reduce the necessary forward-auth middlewares and remove a lot of almost-identical code in the gateway. Of course for accessing "off-the-shelf" backend services we still need an appropriate middleware to match the expected header structure on a case-by-case basis.

- All Renku backend service APIs receive the same headers fields in forwarded authenticated API requests.
- All Renku backend service APIs can trust those headers and rely on the fact that JWT signatures have been verified.
- Renku backend service APIs must not be exposed to the internet directly.
- Renku specific headers start with `Renku`, Renku-specific auth-related header start with `Renku-Auth`.
- All other headers (including the standard `Authorization: bearer <keycloak-access-token>` that clients use to authenticate their requests to the API gateway) will be stripped and not be forwarded to the backend service.
- Proposed headers:
```
Renku-Auth-Access-Token: <keycloak access token>
Renku-Auth-Id-Token: <keycloak id token>
Renku-Auth-Git-Credentials: <base64 encoded JSON object containing all GitLab/GitHub access credentials>
```
while the git credentials object would be something like this:
```
renkuAuthGitCredentials = {
  "https://renkulab.io/gitlab": {
    "provider: "GitLab",
    "AuthorizationHeader": "bearer <user gitlab oAuth2 token>"
  },
  "https://dev.renku.ch/gitlab": {
    "provider: "GitLab",
    "AuthorizationHeader": "token <user gitlab access token>"
  },
  "https://github.com":  {
    "provider: "GitHub",
    "AuthorizationHeader": "bearer <user github oauth token>"
  }
}
```
of course initially and as long as one RenkuLab instance can only connect to one GitLab instance, this JSON object would only have one entry.

olevski:
This sounds good to me! I very much agree with making the traefik config file and in general the renku-gateway code shorter.

rokroskar:
I think these headers are fine - what I would really like to see though is that in the auth implementation of the non-JH-based notebooks service this piece is modular. I.e. I want to make sure that "renku" bits and pieces are not hard-coded and can be swapped out for something more generic to make the service more broadly usable.

PR #414, opened 2021-05-03
feat: add renku headers to nb service requests

ableuler:
This PR adds the header
`Renku-Gitlab-Accesstoken: <some-token>`
together with some other Renku specific header to authenticated requests to the notebooks service. Note that this is only a temporary solution as the expected header names / content will be synchronized between the different renku backend services.

/deploy

### DIFF
--- a/helm-chart/renku-gateway/templates/configmap.yaml
+++ b/helm-chart/renku-gateway/templates/configmap.yaml
@@ -64,7 +64,7 @@ data:
 {{ if eq .Values.global.anonymousSessions.enabled false }}
         [http.routers.notebooks]
           entryPoints = ["http"]
-          Middlewares = ["auth-jupyterhub", "common", "notebooks"]
+          Middlewares = ["auth-jupyterhub", "auth-notebook", "common", "notebooks"]
           Rule = "PathPrefix(`{{ .Values.gatewayServicePrefix | default "/api/" }}notebooks`)"
           Service = "jupyterhub"
 {{ end }}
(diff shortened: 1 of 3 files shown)


helm-chart/renku-gateway/templates/configmap.yaml
```diff
@@ -201,6 +201,11 @@ data:
           trustForwardHeader = true
           authResponseHeaders = ["Authorization", "Renku-user-id", "Renku-user-fullname", "Renku-user-email"]
 
+        [http.middlewares.auth-notebook.forwardauth]
+          address = "http://{{ template "gateway.fullname" . }}-auth/?auth=notebook"
+          trustForwardHeader = true
+          authResponseHeaders = ["Renku-gitlab-accesstoken", "Renku-user-id", "Renku-user-fullname", "Renku-user-email"]
```
olevski, line 207: @ableuler I see all of `"Renku-gitlab-accesstoken", "Renku-user-id", "Renku-user-fullname", "Renku-user-email"` in the header inside the requests in the notebooks service. However I do not see just the regular `Renku-User` header. Was this also supposed to be included here? If I include `Renku-User` in the list of headers here it does show up in notebooks.

olevski (changes requested):
Andreas I just have one question and a few nit-picky things. Everything is in order though.

ableuler:
Thanks @olevski and sorry for this rather ugly copy-paste PR. Let's keep the PR open so that the notebook service can be developed further against this branch and in the mean time figure out the proper headers structure we want adopt in the future. I've updated the [issue description](https://github.com/SwissDataScienceCenter/renku-gateway/issues/406#issue-864623751).

helm-chart/renku-gateway/templates/configmap.yaml
ableuler: There's a config to determine which headers will actually taken over from the response of the middleware: https://github.com/SwissDataScienceCenter/renku-gateway/blob/429a4004fc91badabef6fbe83a5151426f6bc045/helm-chart/renku-gateway/templates/configmap.yaml#L207

ableuler: So you'd also have to add `Renku-User` in this list in order to receive it in the notebooks service.

helm-chart/renku-gateway/templates/configmap.yaml
```diff
@@ -320,7 +325,7 @@ data:
         [http.routers.notebooksSecondPassRegular]
           priority=1003
           entryPoints = ["http"]
-          Middlewares = ["common", "notebooks"]
+          Middlewares = ["auth-notebook","common", "notebooks"]
```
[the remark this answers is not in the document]
ableuler: 👍

ableuler:
I've now implemented the header structure suggested in #406.

ableuler:
@olevski I think it would be good to merge this PR. It has worked with the operator-backed session POC and doesn't break any of the current functionality.

olevski (approved):

PR #429, opened 2021-05-28
K8s operator POC

olevski:
_


helm-chart/renku-gateway/templates/configmap.yaml
```diff
@@ -234,6 +218,10 @@ data:
           regex = "^/projects/([^/]*)/graph(.*)"
           replacement = "/projects/$1/events$2"
 
+        [http.middlewares.jupyterhub-compat.ReplacePathRegex]
```
ableuler, line 221: I think we can remove all of those traces now - or am I missing something?

olevski: This is just to "fool" the ui and make things work for sending API requests to the notebook service either at `/jupyterhub/services/notebooks/` or `/api/notebooks/`. If we remove this then the UI needs to also update their routes they send the requests to before we can roll this out. This is just to make things compatible with what we have right now - that is why I called it `jupyterhub-compat`.

olevski: For the sake of keeping things simple and not changing too many things at once. I can add an issue to clean this up after we roll things out.

ableuler (approved):

PR #437, opened 2021-06-07
feat(auth): add anonymous id to cookie

ableuler:
This adds an anonymous id `anon-id` to the cookie. The actual ID is added as `Renku-Auth-Anon-Id` header to notebook server requests, independent on wether the user is authenticated or not. 

caveats:
- once the UI server is ready, such a anonymous ID should be provided by the UI server
- for this to work properly, the UI needs to retry the very first AJAX request which sets the anonymous id in the cookie (a page reload will do as well until this is implemented).

/deploy

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -21,6 +21,7 @@ import json
 import logging
 import os
 import re
+import secrets
 import sys
 
 import jwt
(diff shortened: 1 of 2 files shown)


ableuler:
> for this to work properly, the UI needs to retry the very first AJAX request which sets the anonymous id in the cookie (a page reload will do as well until this is implemented).

after chatting with @lorenzo-cavazzi we realized that this is not even necessary and that I just needed to fix the format of the 401 response when setting the anon-id cookie.

olevski (approved):

PR #471, opened 2021-09-24
fix: prevent exception when using keycloak access tokens

lorenzo-cavazzi:
This fixes an exception happening while authenticating with a keycloak access token instead of a session cookie.
It also updates the debugger to use the new debugpy library instead of the deprecated ptvsd (I was getting weird errors on VScode with that).

/deploy
re SwissDataScienceCenter/renku-ui#1337


app/__init__.py
```diff
@@ -170,17 +170,23 @@ def auth():
                 "error": "authentication",
                 "message": "can't refresh access token",
                 "target": auth_arg,
+                "exception": str(error),
             }
             return Response(json.dumps(message), status=401)
         raise
     # TODO: fix bare except
```
ableuler, line 177: You can remove the #noqa and the #TODO here as you just fixed the bare except.

app/auth/web.py
```diff
@@ -66,7 +66,9 @@ def get_valid_token(headers):
         return None
 
     keycloak_oidc_client = current_app.store.get_oauth_client(redis_key)
-    return keycloak_oidc_client.access_token
+    if hasattr(keycloak_oidc_client, "access_token"):
+        return keycloak_oidc_client.access_token
+    return None
```
ableuler, line 71: Should we log an error here before returning `none`? I think getting here means that we receive a cookie or bearer token identifying a logged-in user, but that we don't have a valid gitlab, JH or KC token in the redis store for this user. This shouldn't really happen, and if it happens the client should be informed about it such that it can initiate a re-login flow for the specific backend. We'll need to refactor this further down the road, so for now just logging an error would seem appropriate to me.

ableuler (changes requested):

ableuler:
Thanks for brining the bearer token authentication back to life :)

ableuler (approved):

ableuler:
Note that ci should pass as soon as https://github.com/SwissDataScienceCenter/renku/pull/2142 is merged.

lorenzo-cavazzi:
> Note that ci should pass as soon as [SwissDataScienceCenter/renku#2142](https://github.com/SwissDataScienceCenter/renku/pull/2142) is merged.

Ah great, thanks for mentioning it! I was wondering if this PR introduced an error on the anonymous auth :grin: 
I'll go on and merge it then, unless you think it's better to hold on and re-run the tests later for safety

ableuler:
just hold on for a few more hours - the other PR will be merged today and after that tests should run properly.

Issue #473, opened 2021-09-28
Logging out in Renku UI does not log user out of Gitlab


olevski:
## Description

Logging out from the Renku UI should log the user out of Renkulab and Gitlab. But the user remains logged into Gitlab. If the user logs out only from the UI and then logs back in with another account they can still see only things from the initial account.

## Reproduce

1. log into renkulab.io with account 1
2. log out of renkulab.io
3. check renkulab.io/gitlab -> you can still see account 1 logged in
4. navigate back to renkulab.io -> you are not logged in
5. log in to renkulab.io with account 2 -> now you only see stuff from account 1 both in renkulab.io and in renkulab.io/gitlab

## Context

- renkulab url: https://renkulab.io
- operating system and version: mac os big sur 11.6
- browser and version: Chrome Version 93.0.4577.82 (Official Build) (x86_64)

ableuler:
Moved the issue here as it's related to this line of code: https://github.com/SwissDataScienceCenter/renku-gateway/blob/3977b4a9865108f381543e990c9bf6916b0d6717/app/auth/web.py#L199

ableuler:
It should be possible to specify at the level of the helm chart if a logout from the RenkuLab platform should also trigger a logout from GitLab. For deployments which include a dedicated GitLab instance, triggering GitLab logout is the expected behaviour. In the case of "external" GitLab instances which do not rely on RenkuLabs Keycloak as identity provider, no logout from GitLab should be triggered.

The default in the renku chart is to deploy a dedicated instance, therefore, the default should be to trigger a GitLab logout on RenkuLab logout. 

In the case where we _do_ trigger a logout from GitLab, this logout should log the browser out from GitLab (hence terminating the login session of the browser with GitLab), but _not_ delete the `gitlab_oauth_client` instance (and hence the users gitlab oauth token) from the redis store.

PR #474, opened 2021-09-29
fix: log out from gitlab upon renku logout

mohammad-sdsc:
ATM, we don't log out from Gitlab when logging out from renku after implementing CLI login. CLI only needs gitlab oauth client for authentication and it's fine to log out from gitlab. This PR allows logging out from gitlab upon a renku logout optional, which is enabled by default.

This is deployed on `mohammad.dev.renku.ch`

Fixes: #473


ableuler (commented):
This looks good, thanks a lot. As soon as the ci deployment comes up (after https://github.com/SwissDataScienceCenter/renku/pull/2142 is merged) we can do a quick test and approve the PR.

PR #486, opened 2021-10-28
feat: modify for custom CA certificates

olevski:
Allows the gateway to properly access another service (i.e. namely gitlab and/or keycloak) if that service is using self-signed certificates.

I followed the Gitlab cloud native helm chart and how they tackle this issue when addressing this. See for examples:

https://gitlab.com/gitlab-org/charts/gitlab/-/blob/master/templates/_certificates.tpl
https://gitlab.com/gitlab-org/charts/gitlab/-/blob/master/charts/gitlab/charts/webservice/templates/deployment.yaml#L102

Fixes SwissDataScienceCenter/renku#2322


ableuler (approved):
Thanks for taking care of this!

PR #490, opened 2021-11-05
Add GitLab auth to /direct endpoints

lorenzo-cavazzi:
The route `/api/direct` already exists to bypass the usual `/api/v4` prefix we add to queries going to GitLab.
The UI now requires access to GraphQL ( SwissDataScienceCenter/renku-ui#1529 ) .

We could use this rule, adding the GitLab access token when the user is authenticated.

/deploy


lorenzo-cavazzi:
Superseeded by #491

PR #491, opened 2021-11-11
Add GitLab auth to /direct endpoints

lorenzo-cavazzi:
The route `/api/direct` already exists to bypass the usual `/api/v4` prefix we add to queries going to GitLab.
The UI now requires access to GraphQL ( SwissDataScienceCenter/renku-ui#1529 ) .

We could use this rule, adding the GitLab access token when the user is authenticated.

/deploy

### DIFF
--- a/helm-chart/renku-gateway/templates/configmap.yaml
+++ b/helm-chart/renku-gateway/templates/configmap.yaml
@@ -86,7 +86,7 @@ data:
           # This is to access undocumented APIs in GitLab
           entryPoints = ["http"]
           Middlewares = [
-            "auth-gitlab","direct"
+            "direct"
             {{- if .Values.global.gitlab.urlPrefix -}}
             {{- if and (ne .Values.global.gitlab.urlPrefix "") (ne .Values.global.gitlab.urlPrefix "/") -}}
             ,"gitlabOnly"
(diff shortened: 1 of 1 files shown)


helm-chart/renku-gateway/templates/configmap.yaml
```diff
@@ -86,7 +86,7 @@ data:
           # This is to access undocumented APIs in GitLab
           entryPoints = ["http"]
           Middlewares = [
-            "direct"
+            "auth-gitlab","direct"
```
ableuler, line 89: Are those APIs we're using here still undocumented? Yet they do authentication with the users oauth token? It's a bit scary to rely on this 😬

ableuler (approved):

helm-chart/renku-gateway/templates/configmap.yaml
lorenzo-cavazzi: No, the authentication is for GrpahQL. Maybe we should have a separate route for that?

lorenzo-cavazzi:
@ableuler on a second thought, I feel re-using the `direct` endpoint to allow invoking the GitLab GraphQL endpoint was not a great choice. Also, that was missing the rate limiter.
I created a different route that the client can invoke using `/graphql`

helm-chart/renku-gateway/templates/configmap.yaml
```diff
@@ -95,6 +95,20 @@ data:
           ]
           Rule = "PathPrefix(`{{ .Values.gatewayServicePrefix | default "/api/" }}direct/`)"
           Service = "gitlab"
+        
+        [http.routers.gitlabGraphql]
+          entryPoints = ["http"]
+          Middlewares = [
+            "auth-gitlab","general-ratelimit","api"
+            {{- if .Values.global.gitlab.urlPrefix -}}
```
ableuler, line 103: Hmm, I think we could just make "/" the default URL prefix and define a middleware that works for all url prefixes, no? I know we've used this exact pattern before and you didn't "invent" it for this PR, but it's not exactly pretty...

lorenzo-cavazzi: As discussed offline, it would definitely be nice to harmonize how we handle the GitLab base URL.
Let's move that to a separate issue #493 -- I'll tackle it soon

ableuler (approved):

Issue #493, opened 2021-11-16
Improve how we handle the Gitlab URL in the configmap traefik rules


lorenzo-cavazzi:
We added some extra `if` in the "configmap.yaml" rules to handle different GitLab URLs. It would be nice to harmonize all of them by creating a single middleware to handle the GitLab "base" URL.
We should consider cases like these:
* dev.renku.ch/gitlab
* dev.renku.ch/gitlab/ (mind the tryiling `/`)
* dev.renku.ch/something/somethingElse/gitlab
* gitlab.com

PR #505, opened 2021-12-02
493 gitlab urls

lorenzo-cavazzi:
It turns out that streamlining the logic to handle the GitLab prefix with trailing slashes is not that simple.
The main reason is the impossibility to chain multiple `addPrefix` where one returns an empty string. Also, a `/` is automatically added as the first char if it's not already there.

I also tried with a different approach, using `ReplacePathRegex` instead of multiple `addPrefix`, but that seems to fail on some project paths containing the `/` char -- it's the last version pushed in this PR.

We either need to:
- keep a few `if` here and there, but then we don't really simplify compared to the current solution
- or require to specify the full GitLab URL in the values file.

The last seems (by far) the most reasonable solution. Looking at values files, it seems that having a separate `urlPrefix` is not really useful. And we pass the full GitLab URL also as values local to other charts.
Am I missing some use for the `urlPrefix`, or is it possible it's just a leftover from the old days? Should we consider adding something like `global.gitlab.url` and removing it from the other places?

re #493

### DIFF
--- a/helm-chart/renku-gateway/Chart.yaml
+++ b/helm-chart/renku-gateway/Chart.yaml
@@ -3,4 +3,4 @@ appVersion: '2.0'
 description: A Helm chart for the Renku gateway
 name: renku-gateway
 icon: https://github.com/SwissDataScienceCenter/renku-sphinx-theme/raw/master/renku_sphinx_theme/static/favicon.png
-version: 0.10.2
+version: 0.10.2-n004.h08f8989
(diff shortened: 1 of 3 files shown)


ableuler:
totally agree, global full gitlab URL is the way to go. I think the days when gitlab will be deployed as part of the Renku chart are counted anyway.

rokroskar:
@lorenzo-cavazzi what is the status here?

lorenzo-cavazzi:
Oh, I missed the ping on this!

@rokroskar I could give it another look if it's still relevant -- it was mainly a cleanup on the traefik routing. I'm not sure if we should wait to discuss the recent idea of unifying the values (that might have an impact here) or if we won't go down that road anyway

rokroskar:
I think specifying the full GitLab URL makes sense. Not sure whether the prefix is used or useful still anywhere.

olevski:
@lorenzo-cavazzi I am pretty sure we can fully close this PR?

PR #537, opened 2022-02-08
fix(chart): fully remove parts of non-global redis

olevski:
So these `call-nested` helpers were meant (through some witchery) to call the template of a requirement/child helm chart and pull out the redis release name which is then used in the label to allow client applications to connect to redis.

Now since we rely on a global redis instance we should not use this anymore. There is no local redis anymore.

This was causing the helm chart building/publishing step to fail: https://github.com/SwissDataScienceCenter/renku-gateway/runs/5108179637?check_suite_focus=true

/deploy #persist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,22 +30,20 @@ jobs:
   test-chart:
     needs: test
     runs-on: ubuntu-latest
-    if: "startsWith(github.ref, 'refs/tags/') || github.ref == 'refs/heads/master'"
     steps:
       - uses: actions/checkout@master
       - uses: actions/setup-python@v1
         with:
           python-version: 3.7
       - name: Install helm
-        env:
-          HELM_URL: https://storage.googleapis.com/kubernetes-helm
-          HELM_TGZ: helm-v2.17.0-linux-amd64.tar.gz
-          TEMP_DIR: ${{ runner.temp }}
-        run: ./install_helm.sh
+        uses: azure/setup-helm@v1
+        with:
+          version: '3.8.0'
       - name: Test chart
         run: |
-          PATH=${{ runner.temp }}/linux-amd64/:$PATH
-          helm lint helm-chart/renku-gateway --set oidcClientSecret=oidc-secret,gitlabClientSecret=gitlab-secret,cliClientSecret=cli-secret
+          helm repo add renku https://swissdatasciencecenter.github.io/helm-charts/
+          helm dep update helm-chart/renku-gateway
+          helm lint helm-chart/renku-gateway --set oidcClientSecret=foo,gitlabClientSecret=bar,secretKey=foobar,cliClientSecret=barfoo,global.redis.host=test
       - name: Build chart and images
         run: |
           python -m pip install --upgrade pip pipenv
--- a/helm-chart/renku-gateway/templates/_helpers.tpl
+++ b/helm-chart/renku-gateway/templates/_helpers.tpl
@@ -42,17 +42,25 @@ http
 {{- end -}}
 {{- end -}}
 
+{{/*
+Hack for calling templates in a fake scope (until this is solved https://github.com/helm/helm/issues/3920)
+*/}}
+{{- define "call-nested" }}
+{{- $dot := index . 0 }}
+{{- $subchart := index . 1 }}
+{{- $template := index . 2 }}
+{{- include $template (dict "Chart" (dict "Name" $subchart) "Values" (index $dot.Values $subchart) "Release" $dot.Release "Capabilities" $dot.Capabilities) }}
+{{- end }}
+
+{{/*
+Define the redis hostname either from global values section or from a subchart.
+*/}}
 {{- define "redis.host" -}}
 {{- if .Values.global.redis.host -}}
 # If global hostname for redis is found use that
 {{- .Values.global.redis.host -}}
 {{- else -}}
-{{- if hasKey .Subcharts "redis" -}}
 # If global hostname for redis is not found, then check for redis subchart and then use that
-{{- required "If a subchart for redis is used then redis.fullname should be defined." .Subcharts.redis.fullname -}}
-{{- else -}}
-# There is no redis subchart with fullname AND there is no hostname for redis in the global section - show error message
-{{- required "Either global.redis.host should be defined or a subchart for redis should be used where redis.fullname is available." .Values.global.redis.host -}}
-{{- end -}}
+{{- required "Either global.redis.host should be defined or a subchart for redis should be used where redis.fullname is available." (include "call-nested" (list . "redis" "redis.fullname")) -}}
 {{- end -}}
 {{- end -}}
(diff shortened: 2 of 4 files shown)


olevski:
In addition to the issue with the values extracted from child charts, I also did the following:
- cleaned up the messy defaults/required stuff around the secrets, now the secrets have dummy defaults with comments that indicate that they should be set to proper values for prod
- enabled the tests for the helm chart for every PR

olevski:
... also we should probably not use helm v2 to test the chart.

Panaetius (approved):
thank you!

helm-chart/renku-gateway/templates/_helpers.tpl
```diff
@@ -41,13 +41,3 @@ https
 http
 {{- end -}}
 {{- end -}}
-
-{{/*
-Hack for calling templates in a fake scope (until this is solved https://github.com/helm/helm/issues/3920)
```
ableuler, line 46: All for removing this hack. Last time I checked this was still necessary, however I was using an oder version of helm. The [fix PR](https://github.com/helm/helm/pull/9957) was merged in helm 3.7. Should we let this hack live a little longer or force everybody to use helm 3.7 or later?

.github/workflows/ci.yml
```diff
@@ -30,22 +30,20 @@ jobs:
...
       - uses: actions/setup-python@v1
         with:
           python-version: 3.7
       - name: Install helm
-        env:
-          HELM_URL: https://storage.googleapis.com/kubernetes-helm
-          HELM_TGZ: helm-v2.17.0-linux-amd64.tar.gz
-          TEMP_DIR: ${{ runner.temp }}
-        run: ./install_helm.sh
+        uses: azure/setup-helm@v1
```
ableuler, line 39: much nicer

helm-chart/renku-gateway/values.yaml
```diff
@@ -11,15 +11,20 @@ global:
...
-    ## Client id and client secret of the renku gateway client application
-    ## which is registered in GitLab.
-    # gitlabClientId:
-    # gitlabClientSecret:
+    ## Should be set to a proper value (i.e. by using openssl rand -hex 32) for production.
+    ## Can be set here or as .cliClientSecret outside of .global, the values set ouside of global (if defined) take precedence.
+    cliClientSecret: generate-proper-secret
+    ## Client id and client secret of the renku gateway client application which is registered in GitLab.
+    ## Can be set here or as .gitlabClientId outside of .global, the values set ouside of global (if defined) take precedence.
+    gitlabClientId: renku-ui
```
ableuler, line 24:
```suggestion
    gitlabClientId: #renku
```

helm-chart/renku-gateway/templates/secret.yaml
```diff
@@ -12,5 +12,5 @@ type: Opaque
 data:
   oidcClientSecret: {{ .Values.oidcClientSecret | default .Values.global.gateway.clientSecret | b64enc | quote }}
   gitlabClientSecret: {{ .Values.gitlabClientSecret | default .Values.global.gateway.gitlabClientSecret | b64enc | quote }}
-  gatewaySecret: {{ required "Fill in .Values.secretKey with `openssl rand -hex 32`"  .Values.secretKey | b64enc | quote }}
```
ableuler, line 15: The required flag is actually nice because it makes for more useful error messages. I'd even like to use it for the other secrets in this file too, but afaik this is not possible when you use the `| default` pattern, right?

helm-chart/renku-gateway/templates/_helpers.tpl
ableuler: Ah, I see that this hack fails linting of this chart on its own (ie in absence of renku which brings redis as dependency...). This raises the more fundamental question if this chart can be deployed standalone. If yes, we have to include redis as dependency. If no, it should probably be part of renku... 🤔

ableuler (changes requested):
Thanks for taking the initiative of cleaning things up. There's still a few things that we need to figure out though.

ableuler:
> ... also we should probably not use helm v2 to test the chart.

🙈

helm-chart/renku-gateway/templates/_helpers.tpl
ableuler: What I don't want to is to force a user to specifically define the redis host in the global values. We have to be able to infer it automatically for renku deployments which include redis.

olevski: @ableuler the user already has to define all the other values related to redis in the global portion anyhow. But we can resolve the host issue by using this expression for the host: `{{ .Values.global.redis.host | default (printf "%s-redis" .Release.Name) }}`. This works properly when left to default and paired with renku. The CI deployment for this PR succeeded with this change.

helm-chart/renku-gateway/templates/secret.yaml
olevski: Yeah you cannot have both `required` and `default`. I will switch to having `required` messages everywhere.

helm-chart/renku-gateway/templates/_helpers.tpl
olevski: ... and I tested that I could launch a session.

helm-chart/renku-gateway/templates/secret.yaml
ableuler: no, I don't think we can do that. we need the default to be set by the global values.

olevski: Ok sounds good. I will leave the `default` statements, remove the default secrets I added to the values file and modify the lint command so that it does not fail.

helm-chart/renku-gateway/templates/_helpers.tpl
ableuler: `.Values.global.redis.clientLabel` might be useful anyway as it could be very likely that an externally deployed redis instance requires all potential client pods to have some specific label.

olevski:
@ableuler this is ready now.

helm-chart/renku-gateway/templates/_helpers.tpl
```diff
@@ -51,3 +51,16 @@ Hack for calling templates in a fake scope (until this is solved https://github.
 {{- $template := index . 2 }}
 {{- include $template (dict "Chart" (dict "Name" $subchart) "Values" (index $dot.Values $subchart) "Release" $dot.Release "Capabilities" $dot.Capabilities) }}
 {{- end }}
+
+{{/*
```
ableuler, line 55: ah nice, I like the idea to put this into a separate helper.

ableuler (approved):

helm-chart/renku-gateway/templates/_helpers.tpl
olevski: Yeah seeing that `call-nested` template in a few places scares me a bit.

PR #552, opened 2022-02-16
fix(chart)!: modify values for global redis

olevski:
/deploy renku=fix-global-redis-values renku-ui=use-global-redis #persist

### DIFF
--- a/helm-chart/renku-gateway/templates/deployment.yaml
+++ b/helm-chart/renku-gateway/templates/deployment.yaml
@@ -97,7 +97,7 @@ spec:
         app: {{ template "gateway.name" . }}-auth
         release: {{ .Release.Name }}
         # The label below enables the gateway to connect to redis
-        {{ .Values.global.redis.clientLabel | default (printf "%s-%s" (include "redis.host" .) "client") | nindent 8 }}: "true"
+        {{ .Values.global.redis.clientLabel | toYaml | nindent 8 }}
     spec:
       initContainers:
         {{- include "certificates.initContainer" . | nindent 8 }}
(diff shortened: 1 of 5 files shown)


helm-chart/renku-gateway/values.yaml
```diff
@@ -61,14 +61,18 @@ global:
...
+      ## Set to true if redis host/port point to a redis sentinel.
+      enabled: true
+      masterSet: mymaster
     dbIndex:
-      gateway:
+      gateway: "0"
+    host: renku-redis
+    port: 26379
+    clientLabel: 
+      renku-redis-host: "true"
```
ableuler, line 73: should this be
```suggestion
      renku-redis-client: "true"
```
?

helm-chart/renku-gateway/values.yaml
```diff
@@ -61,14 +61,18 @@ global:
   ## Specify the information required to connect to a redis instance.
   ## All the values below are required.
   redis:
-    ## Host is optional if a subchart for redis is used, otherwise it needs to be provided below.
-    # host: 
-    ## The label used to enable the network policy that allows the gateway to connect to redis.
-    ## If not defined below, it will default to <redis hostname>-client.
-    # clientLabel:
-    isSentinel:
+    sentinel:
```
ableuler, line 64: We're not really consistent on this, but I'm wondering if we should just comment all the global values that we're expecting to be set on the renku chart in a standard deployment? After all, what's set here will have no effect whatsoever.

helm-chart/renku-gateway/templates/deployment.yaml
```diff
@@ -137,11 +137,15 @@ spec:
             - name: GATEWAY_SERVICE_PREFIX
               value: {{ .Values.gatewayServicePrefix | default "/api/" | quote }}
             - name: REDIS_HOST
-              value: {{ template "redis.host" . }}
+              value: {{ .Values.global.redis.host | quote }}
             - name: REDIS_IS_SENTINEL
-              value: {{ .Values.global.redis.isSentinel | quote }}
+              value: {{ .Values.global.redis.sentinel.enabled | quote }}
+            - name: REDIS_MASTER_SET
```
ableuler, line 143: Is this env variable picked up at all?

ableuler (changes requested):
Thanks for the makeover, just a few minor questions

helm-chart/renku-gateway/values.yaml
olevski: @ableuler sure I can comment all of this out. And this will be consistent across all the other repos that use the global redis.

olevski: At first I thought sure we can comment out the global values in the component repos. But then the helm chart linting fails. I think it is better to leave it uncommented than to have to specify a bunch of values when you run helm lint.

helm-chart/renku-gateway/templates/deployment.yaml
olevski: It was meant to be and now it is.

helm-chart/renku-gateway/values.yaml
ableuler: you're right, didn't really think about the linting

ableuler (approved):
👍

PR #556, opened 2022-02-25
refactor: clean up traefik rules

ableuler:
Doesn't do yet what it promises in the branch name (ie convert the toml to yaml) but it already cleans up the helm template for the traefik rules quite a bit and can thus be merged "as is". Note that #559 should be reviewed and merged first.

/deploy #persist

### DIFF
--- a/helm-chart/renku-gateway/templates/configmap.yaml
+++ b/helm-chart/renku-gateway/templates/configmap.yaml
@@ -49,65 +49,44 @@ data:
 
         [http.routers.gateway]
           entryPoints = ["http"]
-          Rule = "PathPrefix(`{{ .Values.gatewayServicePrefix | default "/api/" }}auth`)"
+          Rule = "PathPrefix(`{{ printf "/%s/auth" .Values.gatewayServicePrefix | clean }}`)"
           Service = "gateway"
 
         [http.routers.notebooks]
           entryPoints = ["http"]
           Middlewares = ["auth-notebook", "common"]
-          Rule = "PathPrefix(`{{ .Values.gatewayServicePrefix | default "/api/" }}notebooks`)"
-          Service = "notebooks"
-
-        [http.routers.jupyterhub-compat]
-          entryPoints = ["http"]
-          Middlewares = ["jupyterhub-compat", "auth-notebook", "common"]
-          Rule = "PathPrefix(`/jupyterhub/services/notebooks`)"
+          Rule = "PathPrefix(`{{ printf "/%s/notebooks" .Values.gatewayServicePrefix | clean }}`)"
           Service = "notebooks"
 
         [http.routers.webhooks]
           entryPoints = ["http"]
           Middlewares = ["auth-gitlab", "common", "webhooks"]
-          Rule = "Path(`{{ .Values.gatewayServicePrefix | default "/api/" }}projects/{project-id}/graph/webhooks{endpoint:(.*)}`)"
+          Rule = "Path(`{{ printf "/%s/projects/{project-id}/graph/webhooks{endpoint:(.*)}" .Values.gatewayServicePrefix | clean }}`)"
           Service = "webhooks"
 
         [http.routers.graphstatus]
           entryPoints = ["http"]
           Middlewares = ["auth-gitlab", "common", "graphstatus"]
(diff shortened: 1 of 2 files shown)


helm-chart/renku-gateway/templates/configmap.yaml
```diff
@@ -212,11 +184,8 @@ data:
         [http.middlewares.auth-notebook.forwardauth]
           address = "http://{{ template "gateway.fullname" . }}-auth/?auth=notebook"
           trustForwardHeader = true
-    {{ if .Values.global.anonymousSessions.enabled }}
```
ableuler, line 215: If that header isn't included in the response from renku-notebooks, that doesn't matter, the code should not fail.

helm-chart/renku-gateway/templates/configmap.yaml
```diff
@@ -49,65 +49,44 @@ data:
 
         [http.routers.gateway]
           entryPoints = ["http"]
-          Rule = "PathPrefix(`{{ .Values.gatewayServicePrefix | default "/api/" }}auth`)"
```
ableuler, line 52: Default has been moved out of the templates to values.yaml, `clean` is used to clean paths, eg `/some//things/ -> /some/things` or ``// -> `.

helm-chart/renku-gateway/templates/configmap.yaml
```diff
@@ -117,26 +96,26 @@ data:
...
           Middlewares = ["auth-gitlab", "common", "gitlabApi"]
-          Rule = "PathPrefix(`{{ .Values.gatewayServicePrefix | default "/api/" }}`)"
+          Rule = "PathPrefix(`{{ printf "/%s/" .Values.gatewayServicePrefix | clean }}`)"
           Service = "gitlab"
 
         [http.routers.gitlabRenkuCli]
           # Access to gitlab server from a git repo with CLI token
           entryPoints = ["http"]
           Middlewares = ["auth-cliGitlab", "common", "cliGitlab"]
-          Rule = "PathPrefix(`{{ .Values.global.gitlab.cliUrlPrefix | default "/repos/" }}`)"
```
ableuler, line 127: Wasn't used or referenced anywhere, hard-coded for the time being.

helm-chart/renku-gateway/templates/configmap.yaml
```diff
@@ -226,10 +195,6 @@ data:
           regex = "^/projects/([^/]*)/graph(.*)"
           replacement = "/projects/$1/events$2"
 
-        [http.middlewares.jupyterhub-compat.ReplacePathRegex]
```
ableuler, line 229: Move to Amalthea has been successful, not necessary anymore.

helm-chart/renku-gateway/templates/configmap.yaml
```diff
@@ -263,35 +228,35 @@ data:
...
           [[http.services.gitlab.LoadBalancer.servers]]
-            url = {{ .Values.gitlabUrl | default (printf "%s://%s/gitlab" (include "gateway.protocol" .) .Values.global.renku.domain) | quote }}
+            url = "{{ .Values.gitlabUrl | default (printf "%s://%s/gitlab" (include "gateway.protocol" .) .Values.global.renku.domain) }}""
             weight = 1
 
         [http.services.notebooks.LoadBalancer]
           method = "drr"
           passHostHeader = false
           [[http.services.notebooks.LoadBalancer.servers]]
-            url = {{ .Values.notebooks.hostname | default (printf "http://%s-notebooks" .Release.Name )  | quote }}
```
ableuler, line 273: All those local variables like `.Values.notebooks.hostname` aren't documented or used anywhere, let's just go with the default service names at the moment.

olevski (approved):
LGTM. Definitely easier to follow.

Do you think it will be easier to convert the traefik config to yaml? If so I think we should do this in a subsequent PR.

ableuler:
> LGTM. Definitely easier to follow.
> 
> Do you think it will be easier to convert the traefik config to yaml? If so I think we should do this in a subsequent PR.

Yeah, that was the primary intention, hence the branch naming, then I realised that it's easier to clean up first.

olevski (approved):

PR #559, opened 2022-03-01
chore: fix telepresence dev workflow for v2

ableuler:
The readme edits are very scarce at this point, has to be revisited later during the refactoring again. However, I'd like to merge this as-is to prevent on gigantic refactoring PR from developing.


telepresence-configure.sh
```diff
@@ -0,0 +1,15 @@
...
+FLASK_EXECUTABLE=`pipenv --venv`/bin/flask
+
+echo -e "\nTo start/restart the flask development server, run the following command (which was also copied into your pasteboard):\n"
+echo -e "SSL_CERT_FILE=$TELEPRESENCE_ROOT$SSL_CERT_FILE \\
+REQUESTS_CA_BUNDLE=$TELEPRESENCE_ROOT$REQUESTS_CA_BUNDLE \\
+FLASK_DEBUG=1 \\
+FLASK_APP=app:app \\
+HOST_NAME=$HOST_NAME \\
+OAUTHLIB_INSECURE_TRANSPORT=1 \\
+${FLASK_EXECUTABLE} run" | tee >(pbcopy)
```
olevski, line 12: A quick google of `pbcopy` indicates it is a mac-only thing. So I think we should remove it from here. Manually copying and pasting the command is still ok.

README.rst
```diff
@@ -28,51 +23,36 @@
...
+all requests to the gateway-auth service deployed in the cluster to a flask development 
+server running on your local machine (with hot reloading, etc). You can now use your 
+favourite IDE and develop the component completely locally. 
 
-So what is happening here? The command :code:`make dev` launches telepresence which
-swaps the renku-gateway service in your minikube deployment for a locally running version of
-the gateway served by a flask development server. This gives you live updates on code change
-in a minikube deployment!
+.. _telepresence: https://www.telepresence.io/docs/v2.4/quick-start/
+.. _tele-troubleshooting: https://www.telepresence.io/docs/latest/troubleshooting/
```
olevski, line 54: Here somewhere or maybe even better as a message from the scripts used to launch a telepresence sessions we should put some info on how to clean things up.

I.e. until you run `telepresence leave XXX` your Renku deployment does not go back to normal.

Also it seems that XXX = <service name>-<namespace> but we should double check.

olevski (changes requested):
A few minor details. All else works. Installing telepresence was a bit of a pain but that is because I did not read the docs carefully enough. The troubleshooting section for macs is a must.

telepresence-configure.sh
olevski: or we can leave this in as long as it does not crash the whole script completely on linux and also indicate in the instructions that you can skip the manual copy step only on mac

README.rst
ableuler: If you provide a command with the intercept, the intercept will be terminated as soon as that command exits. I added a comment about this to the readme.

ableuler:
> A few minor details. All else works. Installing telepresence was a bit of a pain but that is because I did not read the docs carefully enough. The troubleshooting section for macs is a must.

Yeah, indeed. I think installing is when I first typed `telepresence alternatives` into my browser search bar 😅. Adapted the PR to your suggestions.

olevski (approved):

PR #561, opened 2022-03-08
refactor: use official traefik chart

ableuler:
/deploy renku=ableuler-patch-1  #persist


aledegano (approved):
LGTM!
I have tested launching an user-session in the CI-deployment and all looks good.

PR #563, opened 2022-03-10
Add middlewares CORS.

aledegano:
/deploy extra-values=gateway.allowOrigin="https://example.com" #persist

### DIFF
--- a/.github/workflows/acceptance-tests.yaml
+++ b/.github/workflows/acceptance-tests.yaml
@@ -28,6 +28,7 @@ jobs:
       renku-ui: ${{ steps.deploy-comment.outputs.renku-ui}}
       test-enabled: ${{ steps.deploy-comment.outputs.test-enabled}}
       persist: ${{ steps.deploy-comment.outputs.persist}}
+      extra-values: ${{ steps.deploy-comment.outputs.extra-values}}
     steps:
       - id: deploy-comment
         uses: SwissDataScienceCenter/renku-actions/check-pr-description@v0.5.0
@@ -65,6 +66,7 @@ jobs:
           renku_graph: "${{ needs.check-deploy.outputs.renku-graph }}"
           renku_notebooks: "${{ needs.check-deploy.outputs.renku-notebooks }}"
           renku_ui: "${{ needs.check-deploy.outputs.renku-ui }}"
+          extra_values: "${{ needs.check-deploy.outputs.extra-values }}"
       - name: Check existing renkubot comment
         uses: peter-evans/find-comment@v1
         id: findcomment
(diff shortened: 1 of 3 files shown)


.github/workflows/acceptance-tests.yaml
```diff
@@ -28,6 +28,7 @@ jobs:
       renku-ui: ${{ steps.deploy-comment.outputs.renku-ui}}
       test-enabled: ${{ steps.deploy-comment.outputs.test-enabled}}
       persist: ${{ steps.deploy-comment.outputs.persist}}
+      extra-values: ${{ steps.deploy-comment.outputs.extra-values}}
```
ableuler, line 31: Added this for the extra-values syntax to actually work.

helm-chart/renku-gateway/values.yaml
```diff
@@ -15,6 +15,11 @@ global:
   gitlab:
     urlPrefix: /gitlab
   gateway:
+    ## Optional list of allowed sources for Cross-Origin Resource Sharing.
+    ## https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
+    # allowOrigin:
+    # - example.com
```
ableuler, line 21: Browsers include the schema in the origin header. I've found that when omitting the schema here, traefik will not include the necessary header in the response, so I'd update the example accordingly.

helm-chart/renku-gateway/values.yaml
```diff
@@ -15,6 +15,11 @@ global:
   gitlab:
     urlPrefix: /gitlab
   gateway:
+    ## Optional list of allowed sources for Cross-Origin Resource Sharing.
```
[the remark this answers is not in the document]
aledegano: ok

helm-chart/renku-gateway/values.yaml
aledegano: I don't fully understand what you are asking. Are you just suggesting to add to the comment that "Providing no origins will result in Traefik not including the origin header altogheter"?

ableuler: Sorry for the clumsy wording. Since browsers include the schema in the origin header and traefik wants the origin header to exactly match one of the allowed origins, using examples like `https://example.com` and `http://foo.example.com` in the comment would be more useful.

aledegano: Yes, you're right, the example should include the protocol. I'll change it now!

ableuler: sorry, protocol, not schema 🤦

ableuler (approved):

PR #576, opened 2022-06-24
chore: prep 0.14.0 release

olevski:
New release of traefik. The most meaningful change is the usage of a helm chart for traefik. This helm chart uses a proper security context and runs as non-root.

I also swapped the .rst CHAGELOG file to markdown.

/deploy #persist renku=ableuler-patch-1


olevski:
We cannot make a release with the traefik helm chart yet.

This  will break any deployment that uses self-signed certificates. Trying to figure out how we can add self-signed certs to the traefik helm chart.

PR #582, opened 2022-07-15
chore: run CA cert injection as non-root

olevski:
The latest certificates image and chart can run as non-root.

This is safer than what we had before.

Also the chart and image tags are not tracking each other. This is on purpose. 

/deploy #persist


lorenzo-cavazzi (approved):

olevski:
The tests are stuck on some recent flaky or problematic test. Testing a few things by hand shows no issues. The gateway properly comes up with the new certificates and functions as expected. The failing tests are not from this change.

PR #586, opened 2022-09-01
chore: prepare release 0.15.0

lorenzo-cavazzi:
Prepare release 0.15.0
/deploy #persist


lorenzo-cavazzi:
The offending part was probably `poetry config virtualenvs.create false` -- see this comment as a reference https://github.com/python-poetry/poetry/issues/6288#issuecomment-1233204293 .
An update to the poetry lock file won't harm anyway, I would leave that in.

olevski:
Superseded by #591

PR #587, opened 2022-09-15
Revert anon id changes and update libraries

lorenzo-cavazzi:
/deploy #persist renku-ui=master

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,24 @@
-FROM python:3.7-slim
+FROM python:3.7-slim as base
+RUN apt-get update && \
+    apt-get install -y curl tini && \
+    rm -rf /var/lib/apt/lists/* && \
+    groupadd -g 1000 renku && \
+    useradd -u 1000 -g 1000 -m renku
+WORKDIR /home/renku/renku-gateway
 
-RUN pip install --upgrade pip==22.2.2 && \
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,13 +1,12 @@
...
-RUN pip install --upgrade pip==22.1.2 && \
+RUN pip install --upgrade pip==22.2.2 && \
     pip install poetry
 
 COPY pyproject.toml poetry.lock /code/
 WORKDIR /code
 
-RUN poetry config virtualenvs.create false && \
-    poetry install
+RUN poetry install
```
olevski, line 9: Since you removed this `poetry config virtualenvs.create false` poetry will create a new virtual env for the packages and not install in the system's python. I tried installing in the system python and things failed with the upgrade of pip.

So to fix this either downgrade pip back to what it was.

Or add `poetry run` to the `ENTRYPOINT` further down in the Dockerfile.

olevski: Currently the gateway auth service does not even start.

olevski:
@lorenzo-cavazzi I will convert this to draft. Not sure we should really be setting anonymous ids in two places.

lorenzo-cavazzi:
I guess this can be closed. @olevski feel free to re-open if this PR is still useful

PR #596, opened 2022-10-17
feat: redirect /gitlab to external GitLab url

wesjdj:
If external GitLab is used, redirect /gitlab to the external GitLab URL, including the trailing path content trailing if it exists.

This should be merged at the same time as SwissDataScienceCenter/renku#2741, which adds a rule to the standard Renku ingress to route traffic to /gitlab to renku-traefik if no Renku-bundled GitLab is deployed

### DIFF
--- a/helm-chart/renku-gateway/templates/configmap.yaml
+++ b/helm-chart/renku-gateway/templates/configmap.yaml
@@ -125,7 +125,7 @@ data:
         {{ if .Values.gitlabUrl }}
         [http.middlewares.gitlabRedirect.redirectRegex]
           regex = "(.*)/gitlab(.*)"
-          replacement = "{{ .Values.gitlabUrl }}${2}"
+          replacement = "{{ trimSuffix "/" .Values.gitlabUrl }}${2}"
         {{ end }}
 
         [http.middlewares.common.chain]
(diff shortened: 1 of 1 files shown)


aledegano:
@olevski @lorenzo-cavazzi can we have your opinion on this change?
This is a needed change for the migration to the Gitlab cloud-native chart that we would like to perform (on dev for starter) soon.
Also the sister to this PR: https://github.com/SwissDataScienceCenter/renku/pull/2741

helm-chart/renku-gateway/templates/configmap.yaml
```diff
@@ -113,6 +121,12 @@ data:
         [http.middlewares.uiRedirect.redirectRegex]
           regex = "http://(.*)/entities/(.*)"
           replacement = "{{(include "gateway.protocol" .)}}://${1}/${2}"
+        
+        {{ if .Values.gitlabUrl }}
+        [http.middlewares.gitlabRedirect.redirectRegex]
+          regex = "(.*)/gitlab(.*)"
+          replacement = "{{ .Values.gitlabUrl }}${2}"
```
olevski, line 128:
```suggestion
          replacement = "{{ trimSuffix "/" .Values.gitlabUrl }}${2}"
```

Not sure if a double slash would be a problem there or not. But I guess better to not take any chances. It could be that someone defines the `gitlabUrl` with a slash at the end in the values file.

And from the regex you used above $2 will be either blank or will start with a `/`.

olevski (changes requested):
Looks good, one minor thing.

helm-chart/renku-gateway/templates/configmap.yaml
wesjdj: Thanks! That is a great way of handling the possible double-slash scenario, which I also encountered when testing.

olevski (approved):

PR #598, opened 2022-10-24
fix: remove trailing slash from redirect links

olevski:
For some reason and setup in our routing navigating to `https://renku-ci-rp-3178.dev.renku.ch/api/auth/gitlab/login/` does not work and results in a 404. But if the trailing slash is removed then everything works.

It is funny because the meta tags and javascript do not contain the trailing slash and things work as expected. 

But if someone does not get redirected automatically and click the link then they will get a 404.

/deploy #persist

### DIFF
--- a/app/templates/redirect.html
+++ b/app/templates/redirect.html
@@ -30,7 +30,7 @@
     </head>
 
     <body>
-        <p>If you are not redirected automatically, follow this <a href="{{redirect_url}}/">link.</a></p>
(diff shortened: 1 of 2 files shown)


olevski:
...found this when trying to setup scripts that play along with our chain of logins so that I can run load testing

mohammad-sdsc (approved):
LGTM, thanks!

lorenzo-cavazzi (approved):

olevski:
There are issues with the acceptance tests. But this change does not have any impact on them. So I will merge this instead of waiting for the acceptance tests to be fixed.

PR #601, opened 2022-11-07
feat: add renku styles

andre-code:
PR to include renky styles


app/templates/assets/fonts/Inter/README.txt
```diff
@@ -0,0 +1,72 @@
...
+  MacOS: https://support.apple.com/en-us/HT201749
+  Linux: https://www.google.com/search?q=how+to+install+a+font+on+gnu%2Blinux
+  Windows: https://support.microsoft.com/en-us/help/314960/how-to-install-or-remove-a-font-in-windows
+
+Android Apps
+
+  https://developers.google.com/fonts/docs/android
+  https://developer.android.com/guide/topics/ui/look-and-feel/downloadable-fonts
+
+License
```
olevski, line 63: Do we have to include this README file?

olevski (commented):
Just one quick question @andre-code.  Thanks for adding this.

olevski:
@andre-code a few more things:
- please include the style in all the templates in `app/templates` they are shown at one point or another
- to serve the static files from flask make a folder called `static` and put it inside the `app` folder
- to get the url for a file in the html templates do the following `<link rel="stylesheet" href="{{ url_for('static', filename='style.css') }}">`
- if you have additional folders under `app/static/` like `app/static/fonts` for example I think it should still work but then you will probably want to include the path under filename - you can play around with this a bit but it should be doable I think

olevski:
I did some more googling if you have a subfolder in your static folder you should be able to reference it like this:
``. This would put the link there for the image located at `app/static/img/myimage.png`.

olevski:
Ok so I will list this here for reference. The gitlab redirect page we see is this:

image


The source code for this page is here:
https://gitlab.com/gitlab-org/gitlab/-/blob/master/app/views/doorkeeper/authorizations/redirect.html.haml

And judging from the code and the discussion in [this open issue](https://gitlab.com/gitlab-org/gitlab/-/issues/15199), it is not possible to add/inject custom css in gitlab.


There is one suggestion in the open issue but it is not very well explained and it seems pretty hacky.

If we really want we can try to inject/rewrite something in this page with a proxy. Since we only care about this page and not all of Gitlab.

But I think for now we can let this go and come back to it at a later time. Also intercepting the oauth flow just to add styling does not sound like a good idea.

olevski (dismissed):

ciyer (approved):
I think it is a good solution to put the assets in the renku-ui repo and just keep the minimal HTML necessary in this repo. And it looks great! 🤩

olevski (approved):

olevski:
The tests on the duplicate PR that was used to test the CI deployment issues all passed except the one about  listing PRs and issues. And when I tested that manually it worked. 

So we should not have any issues here either because this PR is the same. Also I closed that other temporary PR.

PR #602, opened 2022-11-08
feat: core service sticky sessions

olevski:
This is required for https://github.com/SwissDataScienceCenter/renku-python/pull/3178.

I created a really small and simple middleware plugin for traefik: https://github.com/SwissDataScienceCenter/cookiefilter. 

It follows the [traefik example](https://github.com/traefik/plugindemo) and [docs](https://plugins.traefik.io/create) on creating a custom plugin. This is required because the middlewares that come included in traefik cannot conditionally filter cookies. What we currently have in place (and the only thing remotely close to what we need) is [this](https://doc.traefik.io/traefik/middlewares/http/headers/) which simply can remove a specific header. So we currently use this middleware to remove all cookies but we cannot selectively keep some of them.

In order for sticky sessions to work we have to not strip the sticky session cookies. That is why this plugin was needed.

And the only way to add a custom plugin to a traefik helm chart (without switching to the paid version) is to rebuild the image with the plugin in it.

/deploy #persist renku-core=feat-scale-core-service renku-ui=update-renku-version-endpoint extra-values=core.ingress.hosts[0].host=renku-ci-gw-602.dev.renku.ch,core.ingress.tls[0].secretName=renku-ci-gw-602-ch-tls,core.ingress.tls[0].hosts[0]=renku-ci-gw-602.dev.renku.ch

### DIFF
--- a/Dockerfile.traefik
+++ b/Dockerfile.traefik
@@ -1,0 +1,11 @@
+# From: https://traefik.io/blog/using-private-plugins-in-traefik-proxy-2-5/
+FROM alpine:3 as builder
+ARG PLUGIN_MODULE=github.com/SwissDataScienceCenter/cookiefilter
+ARG PLUGIN_GIT_REPO=https://github.com/SwissDataScienceCenter/cookiefilter.git
+ARG PLUGIN_RELEASE_TAG=0.0.1
+RUN apk add --update git && \
+    git clone ${PLUGIN_GIT_REPO} /plugins-local/src/${PLUGIN_MODULE} \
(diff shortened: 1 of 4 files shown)


lorenzo-cavazzi (approved):
Lgtm!

Merging does not necessarily require any of the other pieces referenced on the `deploy` string, right? (I.E. it won't require all the following gateway PR to add the same references and extra values)

It would be nice to update the CI values where needed -- unless that's required by the other PR on renku-core

olevski:
@lorenzo-cavazzi I can update the values file in terraform so that it works with both old and new versions. Then we do not have to worry so much about waiting to merge these things.

PR #609, opened 2022-11-23
fix: handle expiring gitlab tokens

olevski:
Related to: https://github.com/SwissDataScienceCenter/renku-notebooks/pull/1335

The changes are as follows:
- the gateway exposes a new endpoint at /gitlab/exchange
- this endpoint requires a keycloak JWT and when one it is provided it will fetch the gitlab oauth token, refresh it if needed and return it
- in addition to the gitlab token the gateway will accept a valid (i.e. signature-wise) but expired keycloak token and if this happens it will return to the caller a refreshed keycloak JWT.

This is required because the git proxy in the sessions need a way to refresh their gitlab tokens. But we do not feel comfortable adding refresh tokens in the sessions. This may not even be feasible as using the refresh token invalidates it. So in this case the session would get the refresh token from the gateway and if it used it before the gateway eventually does then the gateway's refresh token would be invalid. And the gateway would be "stuck" with an expired access_token and an expired refresh_token that cannot be used for anything. At this hypothetical point the gateway would also have no way (other than involving the user in a new login flow) to acquire a new valid refresh token.

### DIFF
--- a/app/auth/cli_auth.py
+++ b/app/auth/cli_auth.py
@@ -35,7 +35,6 @@ from .utils import (
 
 blueprint = Blueprint("cli_auth", __name__, url_prefix="/auth/cli")
 
-CLI_SUFFIX = "cli_oauth_client"
 SCOPE = ["profile", "email", "openid"]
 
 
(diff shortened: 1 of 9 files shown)


olevski:
will be replaced by #613 which does not accept expired tokens.

olevski:
superseded by #613

PR #610, opened 2022-11-24
chore: revert "feat: core service sticky sessions (#602)"

olevski:
This reverts commit 6cf91d34dadc3e622e406b7f304bbbc731dce34d.

I am reverting this because there are other more pressing issues that need to be merged in to the gateway and released. These are:
- the ability to handle expiring gitlab tokens  #609 
- the use of properly styled redirect pages #601 

These were opened/occurred right after the original PR for these changes was merged.

I will "revert the revert" bringing these changes back in when the above two PRs are merged and a release is made. The changes that are revert here depend on simultaneous releases from the ui and renku python and some testing. So they cannot be easily deployed - especially because not all PRs are fully reviewed.

/deploy #persist


lorenzo-cavazzi (approved):
:back:

Issue #606, opened 2022-11-15
Gitlab oauth tokens now expire


olevski:
We make an assumption that they do not:
https://github.com/SwissDataScienceCenter/renku-gateway/blob/master/app/auth/gitlab_auth.py#L44

We should make sure that the gateway can refresh oauth tokens from gitlab.

olevski:
Should be addressed by:
- https://github.com/SwissDataScienceCenter/renku-notebooks/pull/1338
- https://github.com/SwissDataScienceCenter/renku-gateway/pull/613

PR #613, opened 2022-11-30
feat: add endpoint for refreshing expired gitlab tokens

olevski:
/deploy #persist renku=master
This adds an endpoint where a valid renku keycloak access token can be used to get a new gitlab access token.

This endpoint is reachable from outside the cluster. It needs to be because the main users are the sessions and the sessions cannot do any requests that target the internal cluster components. Also now because it is possible to run sessions in a separate namespace even if the session could call the other service in the cluster internally it would not work because these services could be in another namespace.

closes https://github.com/SwissDataScienceCenter/renku-gateway/issues/606


app/config.py
```diff
@@ -86,14 +89,16 @@
     "WEBHOOK_SERVICE_HOSTNAME", "http://renku-graph-webhooks-service"
 )
 
-KEYCLOAK_URL = os.environ.get("KEYCLOAK_URL")
-if not KEYCLOAK_URL:
+KEYCLOAK_URL = os.environ.get("KEYCLOAK_URL", HOST_NAME.strip("/") + "/auth")
+if KEYCLOAK_URL:
+    KEYCLOAK_URL = KEYCLOAK_URL.strip("/")
```
lorenzo-cavazzi, line 94: Doesn't this always test `True` since you define a default for `KEYCLOAK_URL` in the previous line?

app/auth/gitlab_auth.py
```diff
@@ -105,3 +108,22 @@ def logout():
...
+        access_token, key_suffix=current_app.config["GL_SUFFIX"]
+    )
+    # NOTE: getting the client from the redis store automatically refreshes if needed
+    gl_oauth_client: RenkuWebApplicationClient = current_app.store.get_oauth_client(
+        redis_key
+    )
+    return jsonify(
+        {
+            "access_token": gl_oauth_client.access_token,
+            "expires_at": gl_oauth_client._expires_at,
```
lorenzo-cavazzi, line 127: I know this doesn't come from you nor this PR, but why do we have the expiry date starting with `_` as if it were supposed to be a private property? Should we change it to `expires_at` in the `RenkuWebApplicationClient` and adapt to the change where it's used?

lorenzo-cavazzi (changes requested):
The code looks good! Just a couple of tiny details that I commented inline.
I'm deploying the PR to test it

app/auth/gitlab_auth.py
olevski: @lorenzo-cavazzi I am hesitant to touch this because `gl_oauth_client` is an instance of `RenkuWebApplicationClient` which is a child of `oauthlib.oauth2.WebApplicationClient`. And the field `_expires_at` comes from the parent class (from the external library). And we already use `_expires_at` in several  places in our code that I did not touch at all. For example, we use it when we refresh access tokens and when we decide if a token expires soon. So I can make a parameter or function that uses `_expires_at` under the hood and returns the same value but is "public". But this is just cosmetics, our dependence on `_expires_at` is still there. And removing this dependence would require a refactor of the code that I do not want to do right now.

olevski: The plan is to fully refactor the gateway eventually and streamline/remove a lot of these problems. I plan on submitting a RFC for this.

lorenzo-cavazzi (approved):
I tried to significantly lower the token duration (1 minute before refreshing, 15 minutes to expire) and tested UI and Session APIs. All worked well.
I hope that is a good enough test in this case. If not, let me know what I could test better. P.S. I didn't try on the CLI

PR #642, opened 2023-04-20
fix: redirect to None on login or logout

olevski:
Fixes a bug where if the query parameter for the redirect is not passed in then the redirect points to `None`.

/deploy #persist


olevski:
Fixes these kind of errors: https://sentry.dev.renku.ch/organizations/sentry/issues/735/?environment=renkulab&project=8&statsPeriod=7d

PR #643, opened 2023-04-20
fix(chart): HPA for reverse proxy

olevski:
The templating for the horizontal pod autoscaler was using a mix of apiVersion v1 and v2.

Now we use v2 and v2beta2 which are idential in terms of what we use from them and this covers any k8s version older than 1.12.

/deploy #persist extra-values=gateway.reverseProxy.autoscaling.enabled=true

### DIFF
--- a/helm-chart/renku-gateway/templates/hpa-revproxy.yaml
+++ b/helm-chart/renku-gateway/templates/hpa-revproxy.yaml
@@ -1,5 +1,9 @@
 {{- if .Values.reverseProxy.autoscaling.enabled }}
+{{- if semverCompare ">=1.23.0-0" .Capabilities.KubeVersion.GitVersion -}}
 apiVersion: autoscaling/v2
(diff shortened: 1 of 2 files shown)


helm-chart/renku-gateway/templates/hpa-revproxy.yaml
```diff
@@ -20,12 +24,16 @@ spec:
     - type: Resource
       resource:
         name: cpu
-        targetAverageUtilization: {{ .Values.reverseProxy.autoscaling.targetCPUUtilizationPercentage }}
+        target:
+          type: Utilization
+          averageUtilization: {{ .Values.reverseProxy.autoscaling.targetCPUUtilizationPercentage }}
     {{- end }}
```
aledegano, line 30: Thanks 🙏

aledegano (approved):

PR #646, opened 2023-05-08
feat: core-service sticky sessions

Panaetius:
merge with https://github.com/SwissDataScienceCenter/renku-python/pull/3178

### DIFF
--- a/cmd/revproxy/config.go
+++ b/cmd/revproxy/config.go
@@ -13,7 +13,6 @@ import (
 )
 
 type renkuServicesConfig struct {
-	Namespace        string   `mapstructure:"renku_services_namespace"`
 	Notebooks        *url.URL `mapstructure:"renku_services_notebooks"`
 	KG               *url.URL `mapstructure:"renku_services_kg"`
 	Webhook          *url.URL `mapstructure:"renku_services_webhook"`
--- a/cmd/revproxy/config.go
+++ b/cmd/revproxy/config.go
@@ -13,14 +13,13 @@ import (
 )
 
 type renkuServicesConfig struct {
-	Namespace       string   `mapstructure:"renku_services_namespace"`
-	Notebooks       *url.URL `mapstructure:"renku_services_notebooks"`
-	KG              *url.URL `mapstructure:"renku_services_kg"`
-	Webhook         *url.URL `mapstructure:"renku_services_webhook"`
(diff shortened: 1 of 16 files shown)


internal/stickysessions/main.go
```diff
@@ -78,19 +83,20 @@ func NewStickySessionBalancer(service string, namespace string, containerPortNam
...
 	endpointStore := NewEndpointStoreFromEndpointSlices(slices, containerPortName)
 
-	cache := endpointInformer.Lister()
-	watcher := endpointInformer.Informer()
 	balancer := StickySessionBalancer{
 		Service:           service,
 		Namespace:         namespace,
 		ContainerPortName: containerPortName,
-		CookieName:        "reverse-proxy-sticky-session",
+		CookieName:        fmt.Sprintf("reverse-proxy-sticky-session-%s", service),
```
olevski, line 99: I know this is still a WIP. But when I was testing stuff out today I saw something that will cause trouble.

Namely we need to set here a `CookiePath` parameter for the cookie. Because without it the cookies created have a Path based on the request that was sent. So you end up getting a different sticky session cookie per path.

This is a shortcoming that was present in the code from before. I just did not see it because I did not have a chance to test it out fully. This is where the `CookiePath` parameter from here should be set in the cookie https://github.com/SwissDataScienceCenter/renku-gateway/blob/master/internal/stickysessions/main.go#L128

olevski (approved):

cmd/revproxy/config.go
```diff
@@ -13,11 +13,13 @@ import (
 )
 
 type renkuServicesConfig struct {
-	Notebooks *url.URL `mapstructure:"renku_services_notebooks"`
-	KG        *url.URL `mapstructure:"renku_services_kg"`
-	Webhook   *url.URL `mapstructure:"renku_services_webhook"`
-	Core      *url.URL `mapstructure:"renku_services_core"`
-	Auth      *url.URL `mapstructure:"renku_services_auth"`
+	Namespace        string   `mapstructure:"renku_services_namespace"`
```
olevski, line 16:
```suggestion
```

olevski (changes requested):

cmd/revproxy/config.go
olevski: This is crashing the helmchart.

olevski (approved):

PR #647, opened 2023-05-09
squashme: fixups

olevski:
@Panaetius these are some minor fixups for your code.

Lets talk about them tomorrow. Most of the changes address problems that were already there in the sticky sessions load balancer implementation that I did.

### DIFF
--- a/helm-chart/renku-gateway/templates/_helpers.tpl
+++ b/helm-chart/renku-gateway/templates/_helpers.tpl
@@ -41,3 +41,34 @@ https
 http
 {{- end -}}
 {{- end -}}
+
+{{/*
+Template core service paths as a comma separated list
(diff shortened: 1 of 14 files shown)


helm-chart/renku-gateway/templates/_helpers.tpl
```diff
@@ -41,3 +41,34 @@ https
 http
 {{- end -}}
 {{- end -}}
+
+{{/*
+Template core service paths as a comma separated list
+*/}}
+{{- define "gateway.core.paths" -}}
+{{- $paths := list -}}
+{{- range $k, $v := .Values.global.core.versions -}}
```
Panaetius, line 50: `range` order is not deterministic for dict keys

PR #654, opened 2023-06-02
feat: add sentry to reverse-proxy

### LINKED ISSUE #645: Add sentry to the reverse proxy

Panaetius:
closes #645


cmd/revproxy/main.go
```diff
@@ -117,13 +132,16 @@ func main() {
...
+	<-quit // Wait for interrupt signal from OS
 	// Start shutting down servers
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
 	if err := e.Shutdown(ctx); err != nil {
 		e.Logger.Fatal(err)
 	}
+	if config.Sentry.Enabled {
+		defer sentry.Flush(2 * time.Second)
+	}
```
olevski, line 146: I think it is nicer if you move this right below the command that initializes Sentry (i.e. line 113). Because as it stands right now if we hit the line that says `e.Logger.Fatal(err)` immediately above then this will never run. I already made a suggestion with the edit.

cmd/revproxy/main.go
```diff
@@ -97,6 +99,19 @@ func setupServer(config revProxyConfig) *echo.Echo {
...
+	if config.Sentry.Enabled {
+		err := sentry.Init(sentry.ClientOptions{
+			Dsn:         config.Sentry.Dsn,
+			Environment: config.Sentry.Environment,
+			SampleRate:  config.Sentry.SampleRate,
+		})
+		if err != nil {
+			log.Printf("sentry.Init: %s", err)
+		}
+	}
```
olevski, line 115:
```suggestion
	    defer sentry.Flush(2 * time.Second)
	}
```

olevski: See my comment below about this suggestion.

olevski (changes requested):

olevski (approved):

PR #655, opened 2023-06-09
fix: logout with new keycloak version

olevski:
/deploy #persist renku=master

Fixes the logout for the update of keycloak.

The reason that this is required is because of this: https://www.keycloak.org/docs/20.0.5/upgrading/index.html#openid-connect-logout

Namely after keycloak version 18 logging out can be done fully only after the user clicks on a confirmation page that is shown by keycloak. So prior to version 18 going to something like https://dev.renku.ch/auth/realms/Renku/protocol/openid-connect/logout would simply log you out from Keycloak and destroy any Keycloak cookies in the browser. But after version 18 going to that page renders a page that asks the user the confirm the logout request. And only after the request is confirm is the keycloak session and cookies fully destroyed.

The only way to bypass the confirmation page is to provide a query parameter with the id_token. 

The gateway redirects to this page in the logout process. And it does not provide the id_token as a query parameter. So it appears as if the logout completed but that is not the case because the Keycloak cookies are still present.

### DIFF
--- a/app/auth/web.py
+++ b/app/auth/web.py
@@ -200,6 +200,13 @@ def user_profile():
 
 @blueprint.route("/logout")
 def logout():
+    redis_key = get_redis_key_from_session(key_suffix=current_app.config["KC_SUFFIX"])
+    client = current_app.store.get_oauth_client(redis_key)
+    id_token = None
+    if client is not None:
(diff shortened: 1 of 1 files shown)


aledegano (approved):
Works for me! 🙏

PR #668, opened 2023-08-16
fix: properly user Redis sentinel client

olevski:
/deploy #persist

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -115,7 +115,7 @@ def setup_redis_client():
         # Set up the redis store for tokens
         current_app.store = OAuthRedis(_client, current_app.config["SECRET_KEY"])
         # We use the same store for sessions.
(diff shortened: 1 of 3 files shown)


olevski:
The reason for this change is that the auth service is not creating a proper (interruption proof) sentinel connection to redis.

This is an easy fix. But we were subclassing the Redis client and initializing it without taking Sentinel into account. Moving to a proper sentinel connection required us to not initialize the Redis client ourselves. So then we could not also subclass the client either.

That is why I overrode the "__getattr__" method on the class. This is called only after a method or attribute that is called normally (through `__getattribute__`) does not exist. This way we can have a proper sentinel connection without subclassing and still hand over anything to methods that are part of the sentinel client.

lorenzo-cavazzi (approved):
Lgtm :+1:

PR #670, opened 2023-08-21
feat: add auth provider for data service

Panaetius:


olevski (approved):
Not sure if this is ready or you will add more changes. But as it currently stands right now it is fine.

PR #673, opened 2023-08-27
fix(proxy): sticky sessions null dereference

olevski:
~This is just a draft, not ready for review yet.~

/deploy #persist

### DIFF
--- a/internal/stickysessions/endpoints.go
+++ b/internal/stickysessions/endpoints.go
@@ -62,20 +62,21 @@ type EndpointStore struct {
 // The items' index fields will be overwritten accordingly.
 func NewEndpointStoreFromEndpointItems(input []EndpointStoreItem, includeNonReady bool) *EndpointStore {
 	endpoints := &EndpointStore{
-		list:  make([]*EndpointStoreItem, len(input)),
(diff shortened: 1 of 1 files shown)


olevski:
The main reason for this bug is that I pre-allocate and pre-populate the array that will hold the endpoint slices. This "pre-populate" step is done by golang automatically with the "null" version of every value. And the null version of a pointer is a null pointer. Since the array contains endpoint slice pointers when the core version is not ready and the reverse proxy is starting up then it will keep 1 or 2 null pointers in the array because it will not replace the elements with the real endpoint slice if the endpoint slice is not pointing to a fully running core service.

Then further in the code it goes to sort the array of endpoint slices pointers and for this it needs to dereference the pointers because it uses a property of the endpoint slice as the sorting criteria. This is where things crash and burn.

The fix was as simple as replacing `make([]*EndpointStoreItem, len(input))` with `make([]*EndpointStoreItem, 0, len(input))`. The first statement allocates and populates, the second only allocates but does not populate.

The other changes here are only cosmetic.

aledegano (approved):
Ah, yes our dear old Null Point Dereference :D
Thanks for the fix!

PR #692, opened 2023-12-14
fix: handle missing gitlab credentials

olevski:
This was completely crashing the gateway.

Because the client that is extracted can be None but `get_git_credentials_header` assumes that this cannot happen and the client that is passed into the function CANNOT be None.

It happened only somtimes but still we should fix it. With the expiring gitlab tokens it can happen more often.

With this not crashing now the notebook service can receive a request from an authenticated user where the git access tokens header is not speicified. I opened another PR on the notebooks to handle this. See: https://github.com/SwissDataScienceCenter/renku-notebooks/pull/1735

/deploy renku=fix-gitlab-token-in-sessions renku-notebooks=fix-handle-missing-gitlab-tokens


olevski:
This is the that can be easily seen in any gateway logs that this fixes:

```
ERROR:root:'NoneType' object has no attribute 'access_token'
[2023-12-14 10:40:00,924] WARNING in __init__: Error while authenticating request, can't refresh access token. Target: notebook
Traceback (most recent call last):
  File "/code/app/__init__.py", line 181, in auth
    headers = auth.process(request, headers)
  File "/code/app/auth/notebook_auth.py", line 75, in process
    [gitlab_oauth_client]
  File "/code/app/auth/notebook_auth.py", line 49, in get_git_credentials_header
    for client in git_oauth_clients
  File "/code/app/auth/notebook_auth.py", line 49, in <dictcomp>
    for client in git_oauth_clients
AttributeError: 'NoneType' object has no attribute 'access_token'
```

aledegano (dismissed):

aledegano (approved):

PR #696, opened 2024-01-17
feat!: login refactor

olevski:
Things that are left to do:
- logout handlers
- refresh tokens when they are retrieved form the db
- endpoint where the notebook service can swap gitlab access tokens for the sessions
- instructions on how to run locally
- CI pipeline and automated tests
- coveralls

/deploy renku=feat-gateway-refactor renku-ui=feat-gateway-refactoring renku-core=fix--use-RS256-for-token-validation #persist


internal/models/session_handler.go
```diff
@@ -0,0 +1,271 @@
...
+type SessionHandler struct {
+	cookieTemplate           func() http.Cookie
+	sessionTTL               time.Duration
+	tokenStore               TokenStore
+	sessionStore             SessionStore
+	createSessionIfMissing   bool
+	recreateSessionIfExpired bool
+	contextKey               string
+	headerKey                string
+	idQueryKey                 string
```
Panaetius, line 26: shouldn't gofmt indent this the same as the other lines?

internal/revproxy/auth.go
```diff
@@ -0,0 +1,257 @@
...
+				return err
+			}
+			return next(c)
+		}
+	}
+}
+
+var notebooksGitlabAccessTokenHandler TokenHandler = func(c echo.Context, accessToken models.OauthToken) error {
+	// NOTE: As long as the token comes from the database we can trust it and do nto have to validate it.
+	// Each service that the request ultimately goes to will also validate before it users the token
```
Panaetius, line 173:
```suggestion
	// NOTE: As long as the token comes from the database we can trust it and do not have to validate it.
	// Each service that the request ultimately goes to will also validate before it uses the token
```

internal/revproxy/proxies.go
```diff
@@ -0,0 +1,61 @@
...
+		Balancer: middleware.NewRoundRobinBalancer([]*middleware.ProxyTarget{
+			{
+				Name: url.String(),
+				URL:  url,
+			}}),
+	}
+	return middleware.ProxyWithConfig(mwconfig)
+}
+
+func registerCoreSvcProxies(ctx context.Context, e *echo.Echo, revproxyConfig *config.RevproxyConfig, mwFuncs ...echo.MiddlewareFunc) {
```
Panaetius, line 37: I think we should add some documentation on why the core svc is different.

internal/stickysessions/endpoints.go
```diff
@@ -86,7 +86,8 @@ func NewEndpointStoreFromEndpointItems(input []EndpointStoreItem, includeNonRead
 func NewEndpointStoreFromEndpointSlices(input []*discoveryV1.EndpointSlice, containerPortName string) *EndpointStore {
 	items := []EndpointStoreItem{}
 	for _, endpointSlice := range input {
-		items = append(items, NewEndpointStoreItems(endpointSlice, containerPortName)...)
+		es := endpointSlice
```
Panaetius, line 89: should add a comment that this is to deal with the weird go loop behavior so it doesn't get accidentally deleted in the future.

internal/tokenrefresher/tokenrefresher.go
```diff
@@ -0,0 +1,170 @@
...
+		if err != nil {
+			slog.Error("GetRefreshToken failed", "error", err)
+			return err
+		}
+
+		myAccessToken, err := tokenStore.GetAccessToken(ctx, expiringTokenID)
+		if err != nil {
+			slog.Error("GetAccessToken failed", "error", err)
+			return err
+		}
```
Panaetius, line 101: should we collect errors here and return them all at once later? otherwise one error early on prevents all other tokens from being refreshed

internal/db/redisadapters.go
```diff
@@ -0,0 +1,504 @@
...
+func (r RedisAdapter) RemoveIDToken(ctx context.Context, idToken models.OauthToken) error {
+	err := r.removeFromIndexExpiringTokens(ctx, idToken)
+	if err != nil {
+		return err
+	}
+
+	return r.rdb.Del(
+		ctx,
+		idTokenPrefix+idToken.ID,
+	).Err()
```
Panaetius, line 213: this could leak tokens in the db if the second call fails but the first one succeded

Panaetius (changes requested):
as a general style suggestion, we have context in quite a few places as `c` but I think it's more common to use `cx` or `ctx`. Especially since we also use `c` for config, sometimes just lines apart.

I know the Go people are all about cryptic 1 character variable names, but lets not go overboard.

internal/models/serializable_int.go
```diff
@@ -0,0 +1,27 @@
...
+
+type SerializableInt int
+
+func (s SerializableInt) MarshalBinary() (data []byte, err error) {
+	return s.MarshalText()
+}
+
+func (s *SerializableInt) UnmarshalBinary(data []byte) error {
+	return s.UnmarshalText(data)
+}
```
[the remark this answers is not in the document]
olevski: @Panaetius maybe I misunderstand what you mean here or I just don't know about some package you have in mind. But BinaryUnmarshaller and BinaryMarshaller are interfaces (see [here](https://pkg.go.dev/encoding#BinaryMarshaler)). And I have to implement them here so that I can store stuff in redis. In this case rather than trying to find some fancy binary representation of the int I just dump it into text. I will swap it with json.Marshal and json.Unmarshal. I suspect they also just dump the int to a string.

internal/models/session.go
```diff
@@ -0,0 +1,304 @@
...
+
+func (s *Session) Expired() bool {
+	return time.Now().UTC().After(s.CreatedAt.Add(s.TTL()))
+}
+
+func (s *Session) TTL() time.Duration {
+	return time.Duration(s.TTLSeconds) * time.Second
+}
+
+func (s *Session) SaveTokens(ctx context.Context, accessToken OauthToken, refreshToken OauthToken, idToken OauthToken, state string) error {
```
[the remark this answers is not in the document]
olevski: No if you see above in the docs for the struct `ProviderIDs` it says that `ProviderIDs` is a mapping between the oauth state parameter and provider IDs. It is used to ensure that a device login or a different unfinished login flow in the same browser is not used to store the tokens from another login flow. With the state parameter I can be certain that I am storing the right tokens in the right session. So `state` is not the provider name. It is the oauth2 state parameter that is sent around with different requests between a client and the identity provider.

internal/models/session_handler.go
olevski: Yes if you remember to run it.

internal/db/redisadapters.go
olevski: So this should not pose a problem if id tokens start accumulating in the database. And if they have to be overwritten then it is also not a problem.

The ideal way to address this is to add `TTL` for the token in redis. And redis will then automatically clean tokens up when they are expired. I think we should do this in a followup PR.

internal/tokenrefresher/tokenrefresher.go
olevski: Yes we def should. Done.

internal/stickysessions/endpoints.go
olevski: Added.

olevski:
> as a general style suggestion, we have context in quite a few places as `c` but I think it's more common to use `cx` or `ctx`. Especially since we also use `c` for config, sometimes just lines apart.
> 
> I know the Go people are all about cryptic 1 character variable names, but lets not go overboard.

@Panaetius the only place I use `c` for anything context related is for `echo.Context` which is different than `context.Context`.  For `context.Context` I always use `ctx`. In the case of `echo.Context` using `c` for it is standard. If you look at the [documentation](https://echo.labstack.com/docs/quick-start#hello-world) on the echo website they use that all the time. It is similar to how the usually http handler interface is like:

```
func(w http.ResponseWriter, r *http.Request)
```

That snippet is from the official go docs on http: https://pkg.go.dev/net/http#hdr-Servers

PR #708, opened 2024-06-14
feat!: do not strip 'api' from the search service path

olevski:
This is a breaking change it requires the search service and UI to adapt to it.


eikek:
Thanks! 👍🏼

Panaetius (approved):

PR #719, opened 2024-09-06
fix: implement logout for login providers

leafty:
Add browser-based logout for Keycloak and Gitlab.

Previously, the logout flow would not end the Keycloak nor the GitLab browser session. This prevented users from switching accounts because their existing Keycloak and GitLab sessions would be re-used when logging in with the renku gateway.

/deploy renku=leafty/fix-logout-providers

### DIFF
--- a/internal/views/templates/logout.html
+++ b/internal/views/templates/logout.html
@@ -28,17 +28,32 @@ limitations under the License
         <p>Logging you out of RenkuLab...</p>
         <div id="logout-pages">
             {{- range $key, $provider := .providers }}
-            <iframe src="{{$provider.logoutURL}}"></iframe>
+            <iframe id="logout-page-{{$key}}" src="{{$provider.logoutURL}}"></iframe>
             {{- end}}
         </div>
         <script type="text/javascript">
             const redirectURL = {{.redirectURL}};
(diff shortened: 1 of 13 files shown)


internal/views/templates/logout.html
```diff
@@ -0,0 +1,66 @@
...
+<html lang="en" >
+    <head>
+        <meta charset="UTF-8">
+        <meta http-equiv="x-ua-compatible" content="ie=edge">
+        <meta name="viewport" content="width=device-width, initial-scale=1">
+        <meta http-equiv="refresh" content="5; url={{.redirectURL}}" />
+        <link rel="stylesheet" href="{{.renkuBaseURL}}/static/public/theme.css">
+        <title>Renku - login</title>
+        <script type="text/javascript">
+            setTimeout(function(){ window.location.href = "{{.redirectURL}}"; }, 20000);
```
Panaetius, line 29: question: shouldn't this be 5000 since the text below says "5 seconds"?

leafty: I simply ported the existing html pages which existed in the python gateway. The refresh should come from the `meta` tag above.

Panaetius (approved):

PR #744, opened 2024-11-05
fix: pass cookies to data services

olevski:
This allows the data service to easily add authentication for anonymous sessions.

### DIFF
--- a/internal/revproxy/main.go
+++ b/internal/revproxy/main.go
@@ -74,7 +74,7 @@ func (r *Revproxy) RegisterHandlers(e *echo.Echo, commonMiddlewares ...echo.Midd
 	e.Group("/api/kg/webhooks", append(commonMiddlewares, gitlabToken, noCookies, stripPrefix("/api/kg/webhooks"), webhookProxy)...)
 	e.Group("/api/datasets", append(commonMiddlewares, noCookies, regexRewrite("^/api(.*)", "/knowledge-graph$1"), kgProxy)...)
 	e.Group("/api/kg", append(commonMiddlewares, gitlabToken, noCookies, regexRewrite("^/api/kg(.*)", "/knowledge-graph$1"), kgProxy)...)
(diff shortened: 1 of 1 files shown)


internal/revproxy/main.go
```diff
@@ -74,7 +74,7 @@ func (r *Revproxy) RegisterHandlers(e *echo.Echo, commonMiddlewares ...echo.Midd
 	e.Group("/api/kg/webhooks", append(commonMiddlewares, gitlabToken, noCookies, stripPrefix("/api/kg/webhooks"), webhookProxy)...)
 	e.Group("/api/datasets", append(commonMiddlewares, noCookies, regexRewrite("^/api(.*)", "/knowledge-graph$1"), kgProxy)...)
 	e.Group("/api/kg", append(commonMiddlewares, gitlabToken, noCookies, regexRewrite("^/api/kg(.*)", "/knowledge-graph$1"), kgProxy)...)
-	e.Group("/api/data", append(commonMiddlewares, renkuAccessToken, dataGitlabAccessToken, notebooksRenkuRefreshToken, notebooksAnonymousID(r.sessions), noCookies, dataServiceProxy)...)
+	e.Group("/api/data", append(commonMiddlewares, renkuAccessToken, dataGitlabAccessToken, notebooksRenkuRefreshToken, notebooksAnonymousID(r.sessions), dataServiceProxy)...)
```
leafty, line 77: Why? The session ID is already passed as the `Renku-Auth-Anon-Id` header.

leafty: I see that there is a change in https://github.com/SwissDataScienceCenter/renku-data-services/pull/506, but it doesn't really explain why switching to piggy-backing on the cookie is desirable.

leafty: OK, though data services does not control the lifecycle of the cookie, which may have drawbacks.

leafty (approved):

PR #750, opened 2024-12-10
feat: replace notebooks with data service

olevski:
This routes all requests that go to the notebooks to the data service.

Tested the full suite of PRs for this in a separate PR on the renku repo. Also tested to make sure that sessions launched the old way still appear to anonymous and registered users.

### DIFF
--- a/internal/revproxy/main.go
+++ b/internal/revproxy/main.go
@@ -42,7 +42,6 @@ func (r *Revproxy) RegisterHandlers(e *echo.Echo, commonMiddlewares ...echo.Midd
 		gitlabProxy = fallbackProxy
 		gitlabProxyHost = setHost(r.config.RenkuBaseURL.Host)
 	}
-	notebooksProxy := proxyFromURL(r.config.RenkuServices.Notebooks)
 	kgProxy := proxyFromURL(r.config.RenkuServices.KG)
 	webhookProxy := proxyFromURL(r.config.RenkuServices.Webhook)
 	keycloakProxy := proxyFromURL(r.config.RenkuServices.Keycloak)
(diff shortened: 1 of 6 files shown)


internal/revproxy/main.go
```diff
@@ -67,7 +66,8 @@ func (r *Revproxy) RegisterHandlers(e *echo.Echo, commonMiddlewares ...echo.Midd
 	sk.GET("/", echo.NotFoundHandler)
 
 	// Routing for Renku services
-	e.Group("/api/notebooks", append(commonMiddlewares, notebooksRenkuAccessToken, notebooksRenkuRefreshToken, notebooksRenkuIDToken, notebooksGitlabAccessToken, notebooksAnonymousID(r.sessions), noCookies, stripPrefix("/api"), notebooksProxy)...)
+	// Notebooks is being routed to data service now
+	e.Group("/api/notebooks", append(commonMiddlewares, notebooksRenkuAccessToken, notebooksRenkuRefreshToken, notebooksRenkuIDToken, notebooksGitlabAccessToken, notebooksAnonymousID(r.sessions), noCookies, regexRewrite("^/api/notebooks(.*)", "/api/data/notebooks$1"), dataServiceProxy)...)
```
leafty, line 70: Should we keep the same middleware stack as for `data-services`?


```suggestion
	e.Group("/api/notebooks", append(commonMiddlewares, renkuAccessToken, dataGitlabAccessToken, notebooksRenkuRefreshToken, notebooksAnonymousID(r.sessions), regexRewrite("^/api/notebooks(.*)", "/api/data/notebooks$1"), dataServiceProxy)...)
```

olevski: I am not sure that the anonymous sessions in the old notebooks code for v1 sessions will work if we use the same middlewares as for v2.

leafty: If it relies on `notebooksAnonymousID`, it should work.

internal/revproxy/main.go
```diff
@@ -56,18 +55,17 @@ func (r *Revproxy) RegisterHandlers(e *echo.Echo, commonMiddlewares ...echo.Midd
 	dataGitlabAccessToken := r.dataGitlabAccessTokenAuth.Middleware()
 	gitlabToken := r.gitlabTokenAuth.Middleware()
 	gitlabCliToken := r.gitlabCliTokenAuth.Middleware()
-	notebooksRenkuAccessToken := r.notebooksRenkuAccessTokenAuth.Middleware()
 	notebooksRenkuRefreshToken := r.notebooksRenkuRefreshTokenAuth.Middleware()
```
leafty, line 58: nitpick: rename as `renkuRefreshToken`.

leafty (approved):
LGTM

internal/revproxy/main.go
olevski: Yes now this has the same middleware as for data service (except for the path rewrite for notebooks which is necessary).

olevski:
Sorry for dismissing your approval @leafty. Nothing really changed. I just did not see that you already approved. And by habit I resolved the comment and requested the review again.

leafty (approved):

PR #766, opened 2025-04-22
fix: set token ttl correctly in redis

Panaetius:


internal/sessions/token_handling.go
```diff
@@ -175,11 +175,6 @@ func (*SessionStore) idTokenKey(tokenID string) string {
 	return IDTokenCtxKey + ":" + tokenID
 }
 
-// getTokenStorageExpiration returns the max session expiration unless the provider is Renku or GitLab, in which case there is no expiration
-func (*SessionStore) getTokenStorageExpiration(tokens models.AuthTokenSet, session models.Session) time.Time {
-	providerID := tokens.AccessToken.ProviderID
-	if providerID == "renku" || providerID == "gitlab" {
-		return time.Time{}
```
leafty, line 182: This change will break renku v1 if I understand it correctly.

leafty: Forever tokens are needed for some v1 functionality.

Panaetius: this is purely for redis expiration/ttl, not the live time of the token itself or anything like that.

leafty: By doing this change, a TTL will be set on Keycloak and GitLab tokens, which some parts of renku v1 expect to be available forever (even if the user does no log in after session expiry). Feel free to run this as an experiment, but this is the reason redis is remembering tokens forever; I did not want to break renku v1 with the gateway refactor.

olevski: So I am just adding this here after our discussion. The main issue is that SSH sessions would be affected if we started to evict tokens. Because for these users do not need to be logged in once they have started.

PR #769, opened 2025-07-14
Add option to not inject gitlab tokens

ciyer:


internal/revproxy/auth.go
```diff
@@ -18,6 +18,25 @@ import (
 
 type AuthOption func(*Auth)
 
+type AuthMiddlewareProvider interface {
+	Middleware() echo.MiddlewareFunc
+}
+
+func NewAuthMiddlewareProvider(ignoreGitlabTokens bool, options ...AuthOption) (AuthMiddlewareProvider, error) {
```
leafty, line 25: If I set `ignoreGitlabTokens` to true on a middleware that has nothing to do with GitLab, it will skip the auth entirely too, so the way the arguments are named here are an anti-pattern.

internal/config/config.yaml
```diff
@@ -25,6 +25,7 @@ sessions:
 revproxy:
   renkuBaseUrl: "https://renkulab.io"
   externalGitlabUrl:
+  ignoreGitlabTokens: false
```
leafty, line 28: Misnomer: use `enableV1Services: false`, or use `skipGitlabTokens: true`

leafty (commented):
Why not write the logic in `internal/revproxy/main.go`, in `RegisterHandlers()` and `initializeAuth()`? It would be much simpler to follow and understand which middlewares are affected by the change of configuration.

When `enableV1Services` is `false`, then you can:
1. Skip initialization of:
    - `coreSvcIdTokenAuth`
    - `dataGitlabAccessTokenAuth`
    - `gitlabTokenAuth`
    - `gitlabCliTokenAuth`
    - `notebooksGitlabAccessTokenAuth`
2. Update how the revproxy is initialized in `RegisterHandlers()`

internal/revproxy/main_test.go
```diff
@@ -161,26 +161,27 @@ func setupTestUpstream(ID string, requestTracker chan<- *http.Request) (*httptes
...
-		requestTracker <- r
-		w.WriteHeader(responseStatus)
-	}))
-	url, err := url.Parse(srv.URL)
-	if err != nil {
-		log.Fatal(err)
-	}
-	return srv, url
-}
+// TODO Unused code. Can it be deleted?
```
leafty, line 164: Probably yes. There may have been a test using this in the past.

PR #772, opened 2025-09-15
feat: disable v1 svcs but keep internal gitlab

olevski:
This makes it so that we can disable v1 services but still keep the internal gitlab working as it used to do with v1. This is needed so that when we turn off v1 services but we keep the internal gitlab around for a few months users are not asked to make the connection with the internal gitlab integration. And we wouldn't need to add an internal gitlab integration.

/deploy renku=release-2.8.0 extra-values=enableV1Services=false,enableInternalGitlab=true

### DIFF
--- a/internal/revproxy/main.go
+++ b/internal/revproxy/main.go
@@ -54,8 +54,13 @@ func (r *Revproxy) RegisterHandlers(e *echo.Echo, commonMiddlewares ...echo.Midd
 	sk := e.Group("/api/data/user/secret_key", commonMiddlewares...)
 	sk.GET("/", echo.NotFoundHandler)
 
+	if r.config.EnableV1Services && !r.config.EnableInternalGitlab {
+		panic("cannot support turnign")
+	}
+
 	// Middlewares and routing is configured depending on `EnableV1Services`
-	if r.config.EnableV1Services {
+	if r.config.EnableV1Services && r.config.EnableInternalGitlab {
+		// This whole branch of else-if should be removed when the Gitlab is retired.
 		// Initialize common authentication middleware
 		coreSvcIdToken := r.coreSvcIdTokenAuth.Middleware()
 		dataGitlabAccessToken := r.dataGitlabAccessTokenAuth.Middleware()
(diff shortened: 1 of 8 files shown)


internal/revproxy/main.go
```diff
@@ -54,8 +54,13 @@ func (r *Revproxy) RegisterHandlers(e *echo.Echo, commonMiddlewares ...echo.Midd
 	sk := e.Group("/api/data/user/secret_key", commonMiddlewares...)
 	sk.GET("/", echo.NotFoundHandler)
 
+	if r.config.EnableV1Services && !r.config.EnableInternalGitlab {
+		panic("cannot support turnign")
+	}
+
 	// Middlewares and routing is configured depending on `EnableV1Services`
-	if r.config.EnableV1Services {
+	if r.config.EnableV1Services && r.config.EnableInternalGitlab {
```
leafty, line 62: I think it would be clearer to have the three cases in this order:
1. `if r.config.EnableV1Services {`
2. `} else if r.config.EnableInternalGitlab {`
3. `} else {`

internal/revproxy/main.go
```diff
@@ -54,8 +54,13 @@ func (r *Revproxy) RegisterHandlers(e *echo.Echo, commonMiddlewares ...echo.Midd
 	sk := e.Group("/api/data/user/secret_key", commonMiddlewares...)
 	sk.GET("/", echo.NotFoundHandler)
 
+	if r.config.EnableV1Services && !r.config.EnableInternalGitlab {
+		panic("cannot support turnign")
+	}
+
 	// Middlewares and routing is configured depending on `EnableV1Services`
```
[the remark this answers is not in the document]
olevski: addressed in https://github.com/SwissDataScienceCenter/renku-gateway/pull/772/commits/54d2db1ede958e5d2971ce222e3ab05c3bbd2b17

internal/revproxy/main.go
olevski: addressed in https://github.com/SwissDataScienceCenter/renku-gateway/pull/772/commits/54d2db1ede958e5d2971ce222e3ab05c3bbd2b17

leafty (approved):

olevski:
Nvm the config library is a bit weird for cases like this. I will just make the default in the helm chart and then the config will pick that up.

olevski:
Tested this full over at https://ci-renku-4175.dev.renku.ch/

PR #855, opened 2026-04-15
build: upgrade to echo v5

leafty:
Upgrade echo to v5.

Details:
* Follow migration changes for echo v4 to v5: https://github.com/labstack/echo/releases/tag/v5.0.0
* Better handling of IP address extraction, see: https://echo.labstack.com/docs/ip-address

/deploy


cmd/gateway/middlewares.go
```diff
@@ -8,22 +8,22 @@ import (
...
 
 var logLevel *slog.LevelVar = new(slog.LevelVar)
 var jsonLogger *slog.Logger = slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{Level: logLevel}))
 var requestLogger echo.MiddlewareFunc = middleware.RequestLoggerWithConfig(middleware.RequestLoggerConfig{
-	LogStatus:    true,
-	LogURI:       true,
-	LogError:     true,
+	LogStatus: true,
+	LogURI:    true,
+	// LogError:     true,
```
leafty:
```suggestion
```

internal/revproxy/main.go
```diff
@@ -39,15 +39,17 @@ func (r *Revproxy) RegisterHandlers(e *echo.Echo, commonMiddlewares ...echo.Midd
...
 	sk := e.Group("/api/data/user/secret_key", commonMiddlewares...)
-	sk.GET("/", echo.NotFoundHandler)
+	sk.RouteNotFound("*", func(c *echo.Context) error { return echo.ErrNotFound })
 
 	// Redirects store middleware
 	if r.redirects != nil {
 		redirectMiddleware := r.redirects.Middleware()
 		redirectPath := path.Join(r.redirects.PathPrefix, ":projectPath")
-		e.Group(redirectPath, append(commonMiddlewares, renkuBaseProxyHost, redirectMiddleware, fallbackProxy)...)
+		e.Group(redirectPath, append(commonMiddlewares, renkuBaseProxyHost, redirectMiddleware, fallbackProxy)...).Any("*", noOpHandler)
```
olevski: Why do the addition of `Any()` here and in all the other places? This will make it so that even requests with a method that is no implemented returns nil. If I understand the new echo docs correctly.

olevski (changes requested):
Thanks @leafty. One thing that I am confused about is the addition of Any().

internal/revproxy/main.go
leafty: In echo v5.1, registering a proxy as a middleware for a group will not make the group respond unless there are routes registered. This is why we register all routes for all methods with `.Any("*")` so that the proxy middleware can respond.

This new behavior has been reverted in 5.3, so I will upgrade to it. You can see https://github.com/labstack/echo/issues/2950 for context.

olevski: Oh wow. Ok thanks. And thanks for reverting.

olevski (approved):

PR #866, opened 2026-04-30
chore: centralized error formatting

olevski:
/deploy

### DIFF
--- a/internal/revproxy/error_handler.go
+++ b/internal/revproxy/error_handler.go
@@ -1,0 +1,63 @@
+package revproxy
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/labstack/echo/v4"
+)
+
+// The same error content as the data services API
+const gwBaseErrorCode int = 6000
+
+type errorContent struct {
+	Code    int    `json:"code"`
+	Message string `json:"message"`
+	Detail  string `json:"detail,omitempty"`
+	TraceId string `json:"trace_id,omitempty"`
+}
+
+type errorResponse struct {
+	Error errorContent `json:"error"`
+}
+
+// Adapted from https://echo.labstack.com/docs/error-handling
+func ErrorHandler(err error, c echo.Context) {
+	if c.Response() != nil && c.Response().Committed {
+		return // response has been already sent to the client by handler or some middleware
+	}
+
+	accept := c.Request().Header.Get("Accept")
+	isHTML := strings.Contains(accept, echo.MIMETextHTML)
+
+	// If the accept header is html then we fall back to the default handler (for now).
+	// If the acceptt header is not html or is blank we return json
+	if isHTML {
+		c.Echo().DefaultHTTPErrorHandler(err, c)
+		return
+	}
+
+	code := http.StatusInternalServerError
+	message := err.Error()
+	var he *echo.HTTPError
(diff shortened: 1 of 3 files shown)


olevski:
To check this:

Open the browser at: https://renku-ci-gw-866.dev.renku.ch/api/fdfd - this will give you the same response as before.
```
{
  "message": "Not Found"
}
```
We can add a nice http page for the errors later. We need to coordinate with the ui for this.

But if you use curl you will get the same format as the data service.
```
 curl https://renku-ci-gw-866.dev.renku.ch/api/fdfd
{"error":{"code":6404,"message":"Not Found"}}
```

olevski:
With using python we also get the non-http response:
```
>>> import requests
>>> res = requests.get("https://renku-ci-gw-866.dev.renku.ch/api/fd")
>>> res.text
'{"error":{"code":6404,"message":"Not Found"}}\n'
```

internal/revproxy/error_handler.go
```diff
@@ -0,0 +1,63 @@
...
+	"errors"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/labstack/echo/v4"
+)
+
+// The same error content as the data services API
+const gwBaseErrorCode int = 6000
```
leafty, line 13: Why 6000? (Do we have a rule for these numbers?)

internal/revproxy/error_handler.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// Adapted from https://echo.labstack.com/docs/error-handling
+func ErrorHandler(err error, c echo.Context) {
+	if c.Response() != nil && c.Response().Committed {
+		return // response has been already sent to the client by handler or some middleware
+	}
+
+	accept := c.Request().Header.Get("Accept")
+	isHTML := strings.Contains(accept, echo.MIMETextHTML)
```
leafty, line 33: Remark: weak content negotiation. But there is no standard library method to at least parse the accept header (not even try to perform content negotiation).
See: https://httpwg.org/specs/rfc9110.html#field.accept

internal/revproxy/error_handler.go
```diff
@@ -0,0 +1,63 @@
...
+		if msg := fmt.Sprintf("%v", he.Message); len(msg) > 0 {
+			message = msg
+		}
+	}
+
+	var cErr error
+	if c.Request().Method == http.MethodHead {
+		cErr = c.NoContent(code)
+	} else {
+		cErr = c.JSON(code, errorResponse{Error: errorContent{Code: gwBaseErrorCode + code, Message: message}})
```
leafty, line 58: It looks like we are not using `Detail` and `TraceId`. Should we open an issue to add them?

internal/revproxy/error_handler.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	if cErr != nil {
+		c.Logger().Error("failed to send error page to client", "error", errors.Join(err, cErr))
```
leafty, line 61: Use `slog.Error()` here. We configure the logger for `slog` but we don't set the `Logger` used by echo.

internal/revproxy/error_handler.go
leafty, line 1: Note: maybe this file should be in `internal/gwerrors/`?

leafty (commented):
Nice, some comments below:

PR #872, opened 2026-06-15
fix: silence broken pipe errors from web socket connections

leafty:
Filter out the following errors before sending to Sentry:
* 404 errors generated within `echo`
* Broken pipe errors from the echo's proxy middleware

Manually checked in the deployment.

<code>
/deploy #notest extra-values=gateway.sentry.enabled=true,gateway.sentry.environment=renku-ci-gw-872,gateway.sentry.dsn=https://b4747210a4b6a37a4ccc33beee2dbea7@o4509039310995456.ingest.de.sentry.io/4509513007693904
</code>


cmd/gateway/main.go
```diff
@@ -77,7 +78,7 @@ func main() {
 		e.Use(func(next echo.HandlerFunc) echo.HandlerFunc {
 			return func(c echo.Context) error {
 				err = next(c)
-				if err != nil {
+				if err != nil && utils.SendErrorToSentry(err) {
```
sambuc: Should this just be 

```
if utils.SendErrorToSentry(err) {
```
?

I mean we are executing the `nil` check two times, and it is the job of `SendErrorToSentry` to decide if we execute the body of the if block.

I am fine with also checking it here, just asking what's the best practice here.

sambuc (dismissed):

cmd/gateway/main.go
leafty: I'll remove the double check.

aledegano (dismissed):

sambuc (approved):

